PR #3, opened 2019-05-16
AP-441 JSON payload definition and schema validation

stephenrichards:
This PR establishes the JSON schema for the payload to submit an assessment.

The JsonSchemaValidator class will validate the payload and report any errors - this
should be called from the relevant controller action.

TODO: The schema is currently located in /config/api/assessment_schema.json.  This should
probably be moved, as we will want to make the schema available for download by clients so
that they can validate their payload before submitting.


config/api/assessment_schema.json
```diff
@@ -0,0 +1,382 @@
+{
+  "id": "#{Rails.root}/config/api/assessment_schema.json",
```
reggieb: I think this should be `"$id"`. See http://json-schema.org/understanding-json-schema/structuring.html#the-id-property

config/api/assessment_schema.json
```diff
@@ -0,0 +1,382 @@
...
+    "meta_data",
+    "applicant",
+    "applicant_income",
+    "applicant_outgoings",
+    "applicant_capital"
+  ],
+  "additionalProperties": false,
+
+
+  "definitions": {
```
reggieb: I think `$id` can be used here too to clean up some of the uses of these core objects. For example:
```
"uuid": {
      "$id": "#uuid",
      "type": "string",
      "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$",
      "description": "has to be a valid uuid"
    }
```
Then where a uuid is needed instead of:
```
"$ref": "#/definitions/uuid"
```
We can do:
```
"$ref": "#uuid"
```
Again see - http://json-schema.org/understanding-json-schema/structuring.html#the-id-property

config/api/assessment_schema.json
```diff
@@ -0,0 +1,382 @@
...
+      "description": "has to be a valid uuid"
+    },
+    "date": {
+      "type": "string",
+      "pattern": "^([12][0-9]{3}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[01]))$",
+      "description": "Date in format YYYY-MM-DD in range 1900-01-01 to 2999-12-31"
+    },
+    "matter_proceeding_type": {
+      "type": "string",
+      "pattern": "^Domestic abuse$",
```
reggieb: Should this be an enum with only one member?

config/api/assessment_schema.json
```diff
@@ -0,0 +1,382 @@
...
+      "description": "The matter proceeding type - the only one currently handled is 'Domestic abuse'"
+    },
+    "applicant_involvement_type": {
+      "type": "string",
+      "enum": ["applicant", "defendent"],
+      "description": "The type of involvement of the applicant in the case.  Currently limited to applicant in the case, or defendant"
```
reggieb: I'd avoid statements like _"Currently limited to applicant in the case, or defendant"_ in descriptions - as the schema is likely to change but keeping these description in step with those changes will be hard. Let the schema define the current limitations.

config/api/assessment_schema.json
```diff
@@ -0,0 +1,382 @@
...
+    },
+    "number_with_two_decimals": {
+      "type": "number",
+      "multipleOf": 0.01,
+      "minimum": 0.01,
+      "description": "Positive number with two decimal places"
+    },
+    "benefit_name": {
+      "description": "Name of benefit received - must be on eof the values listed in the enum below",
```
reggieb: Typo I think. Should be "must be _one of_ the values"

app/services/json_schema_validator.rb
```diff
@@ -0,0 +1,19 @@
+class JsonSchemaValidator
+
+  SCHEMA_PATH = File.join(Rails.root, 'config', 'api', 'assessment_schema.json')
```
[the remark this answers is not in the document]
stephenrichards: Isn't that just another way of doing the same thing?  Is there a particular advantage to one way over the other, or a documented preferred style?

Not trying to be difficult, but I'd prefer to change things only when there is a reason for doing so, not just one developer's preferred style as opposed to another's.

config/api/assessment_schema.json
stephenrichards: The gem uses draft 4, where it's `id` not `$id`

config/api/assessment_schema.json
stephenrichards: Again, not supported in draft 4.

config/api/assessment_schema.json
stephenrichards: yep - we're going to be adding more, so makes sense.

config/api/assessment_schema.json
stephenrichards: God point.

app/services/json_schema_validator.rb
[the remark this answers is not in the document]
reggieb: I thought you may not be aware of `Rails.root.join`.

config/api/assessment_schema.json
reggieb: Fair enough.

app/services/json_schema_validator.rb
```diff
@@ -0,0 +1,18 @@
+class JsonSchemaValidator
```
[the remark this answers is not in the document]
stephenrichards: Nice

Gemfile
```diff
@@ -28,10 +28,15 @@ gem 'bootsnap', '>= 1.1.0', require: false
 # Use Rack CORS for handling Cross-Origin Resource Sharing (CORS), making cross-origin AJAX possible
 # gem 'rack-cors'
 
+gem 'json-schema'
+gem 'rubocop', require: false
+gem 'rubocop-performance'
```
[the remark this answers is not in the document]
stephenrichards: yep

config/api/assessment_schema.json
```diff
@@ -0,0 +1,376 @@
...
+    "number_with_two_decimals": {
+      "type": "number",
+      "multipleOf": 0.01,
+      "minimum": 0.00,
+      "description": "Non-negative number (including zero) with two decimal places"
+    },
+    "benefit_name": {
+      "description": "Name of benefit received - must be one of the values listed in the enum below",
+      "type": "string",
+      "enum": ["child_allowance", "jobseekers_allowance"]
```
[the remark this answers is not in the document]
stephenrichards: agreed

PR #4, opened 2019-05-20
AP-443 Define logging and response

stephenrichards:
This defines the response from an eligibility check request.

Also included in this PR:

 - move the assessment_request schema to public/schemas so that they can be served by an http request
 - write a schema for the response for documentation purposes
 - a spec to check that the assessment_response schema works
 - create a model to record the request and response.  This will be the log of the activity
0e09375

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -1,4 +1,2 @@
 class Assessment < ApplicationRecord
-  validates :remote_ip, presence: true
-  validates :request_payload, presence: true
 end
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -1,0 +1,4 @@
+class Assessment < ApplicationRecord
+  validates :remote_ip, presence: true
+  validates :request_payload, presence: true
+end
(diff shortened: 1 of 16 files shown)


spec/services/json_assessment_response_validator_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    assessment_request_schema = File.read(Rails.root.join('public/schemas/assessment_request.json'))
+    stub_request(:get, "http://localhost:3000/schemas/assessment_request.json").
+      to_return(:body => assessment_request_schema)
+  end
+
+  context 'A valid response' do
+    it 'does not create any errors' do
+      schema_path = Rails.root.join('public/schemas/assessment_response.json').to_s
+      response_hash = AssessmentResponseFixture.ruby_hash
+      payload = JSON.pretty_generate(response_hash)
```
Kalagan: Could these 2 lines be replaced by 
```ruby
payload = AssessmentResponseFixture.pretty_json
```
?

app/models/assessment.rb
```diff
@@ -0,0 +1,4 @@
+class Assessment < ApplicationRecord
+  validates :remote_ip, presence: true
+  validates :request_payload, presence: true
```
Kalagan: Instead of these validations you could set the database columns to be not null

spec/services/json_assessment_response_validator_spec.rb
stephenrichards: yes, of course.

PR #10, opened 2019-05-22
AP-616 Create RecursiveStruct of request and response

stephenrichards:
## Create RecursiveStruct of request and response

[Link to story](https://dsdmoj.atlassian.net/browse/AP-616)

Added `AssessmentParticulars` which is a specialised `RecursiveOpenStruct` to encapsulate the request and the response as well as the `assessment_id` and the `client_reference-id`.

This allows us to create an `AssessmentParticulars` object and then access inner parts of the structures with ruby method calls, e.g. 
```
particulars = AssessmentParticulars.new(assessment)
particulars.response.assesment_id  # => 'b2f0bb7d-e3d2-4829-a0ae-db1016246c58'
particulars.request.applicant.dependents.first.date_of_birth # => '2012-04-03'
particulars.response.details.result = 'eligible'
```

Also:
- removed `Status` model and associated controllers and specs
- added coverage

## Checklist


app/services/assessment_particulars_creation_service.rb
```diff
@@ -0,0 +1,20 @@
...
+      request: JSON.parse(@request_payload),
+      response: {
+        assessment_id: assessment.id,
+        client_reference_id: assessment.client_reference_id,
+        details: {},
+        errors: []
+      }
+    }
+    RecursiveOpenStruct.new(hash, recurse_over_arrays: true)
+  end
```
reggieb: Define getter methods so that you don't have to access instance variables directly:
```ruby
attr_reader :remote_ip, :request_payload
```

Then break down `call` so it isn't doing too much:
```ruby
def call
  RecursiveOpenStruct.new(data, recurse_over_arrays: true)
end

def assessment 
  @assessment ||= Assessment.create!(remote_ip: remote_ip, request_payload: request_payload)
end

def data
  {
    request: JSON.parse(request_payload),
    response: {
      assessment_id: assessment.id,
      client_reference_id: assessment.client_reference_id,
      details: {},
      errors: []
    }
  }
end
```
Then expose the service functionality via a class method - to ensure that items used in the construction aren't shared across calls
```ruby
def self.create_for(*args)
  new(*args).call
end
```
Then to create a new set of Assessment Particulars:
```ruby
AssessmentParticularsCreationService.create_for remote_ip, request_payload
```

spec/services/assessment_particulars_creation_service_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'rails_helper'
+
+RSpec.describe AssessmentParticularsCreationService do
+  let(:request_payload) { AssessmentFixture.json }
+  let(:request_hash) { AssessmentFixture.ruby_hash }
```
reggieb: Why `ruby_hash`? Why not just `hash`

app/models/assessment_particulars.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  private
+
+  def initial_data(assessment)
+    {
+      assessment_id: assessment.id,
+      client_reference_id: assessment.client_reference_id,
+      request: RecursiveOpenStruct.new(JSON.parse(assessment.request_payload), recurse_over_arrays: true),
+      response: RecursiveOpenStruct.new(initial_response_data, recurse_over_arrays: true)
+    }
```
reggieb: If `RecursiveOpenStruct` is recursive, why does it need to be call again on all of the recursions?

Would this have the same end result:

```ruby
    {
      assessment_id: assessment.id,
      client_reference_id: assessment.client_reference_id,
      request: JSON.parse(assessment.request_payload),
      response: initial_response_data
    }
```

app/models/assessment_particulars.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def initial_response_data
+    {
+      details: RecursiveOpenStruct.new({}, recurse_over_arrays: true),
+      errors: []
+    }
```
reggieb: Again, would this work:
```ruby
    {
      details: {},
      errors: []
    }
```

spec/services/assessment_particulars_creation_service_spec.rb
stephenrichards: because that would override Object#hash

reggieb: OK

app/models/assessment_particulars.rb
```diff
@@ -0,0 +1,36 @@
...
+  def respond_to_missing?(method, include_private = false)
+    @data.respond_to?(method, include_private)
+  end
+
+  private
+
+  def valid_missing_method?(method, args)
+    return true if args.size.zero? && !method.match?(/=$/)
+
+    return true if args.size.zero? && method.match?(/=$/)
```
mpw5: The logic here needs amending.

app/controllers/status_controller.rb
```diff
@@ -1,5 +0,0 @@
-class StatusController < ApplicationController
-  def index
-    render json: { alive: true }
```
Kalagan: This endpoint is used as a probe by kubernetes so it should be left in place. (see https://github.com/ministryofjustice/check-financial-eligibility/blob/master/deploy/helm/templates/deployment_web.yaml#L33)

app/models/assessment_particulars.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  private
+
+  def valid_missing_method?(method, args)
+    return true if setter_method?(method, args)
+
+    return true if getter_method?(method, args)
+
+    false
+  end
```
Kalagan: would this simpler alternative work?
```ruby
  def valid_missing_method?(method, args)
    setter_method?(method, args) || getter_method?(method, args)
  end
```

app/controllers/status_controller.rb
Kalagan: It seems to be 100% now 🙂

PR #11, opened 2019-05-22
AP-617 Adds a store that returns CCMS thresholds

reggieb:
[Jira AP-617](https://dsdmoj.atlassian.net/browse/AP-617)

Creates a generic YAML store class and then uses that to define `CcmsThreshold`. The thresholds are then set in `config/thresholds.yml` in the section `ccms`.

Current behavior allows:
```
2.5.3 :001 > CcmsThreshold.value :capital_upper
 => 10000 
2.5.3 :002 > CcmsThreshold.value :gross_income_upper
 => {:children=>{:one=>1, :two=>3, :many=>7}} 
2.5.3 :003 > CcmsThreshold.value :foo
Traceback (most recent call last):
        3: from (irb):3
        2: from app/models/yaml_store.rb:24:in `value'
        1: from app/models/yaml_store.rb:34:in `value'
YamlStore::KeyNotRecognisedError (key 'foo' not set)

```
_See update to behaviour below_


reggieb:
I've rebuild this solution so that it is now:

- No longer uses CCMS in the name.
- Allows thresholds to be defined for a date range.

The new functionality looks like this:
```
2.5.3 :001 > Threshold.value_for :capital_lower
 => 1 
2.5.3 :002 > Threshold.value_for :capital_lower, at: 1.year.ago
 => 10 
2.5.3 :003 > Threshold.value_for :capital_lower, at: 2.year.ago
 => 10 
2.5.3 :004 > Threshold.value_for :capital_lower, at: 1.year.from_now
 => 5 
2.5.3 :005 > Threshold.value_for :gross_income_upper
 => {:children=>{:one=>1, :two=>3, :three=>5, :many=>7}} 
```

Note that the 2 year ago is before the date set for the first set of thresholds. I've coded it so the oldest thresholds are returned in this case. I could change that to raise an error.

I cannot see a better way of defining `:gross_income_upper`. If you could describe one that is less "awkward" I could change it to behave that way.

PR #12, opened 2019-05-23
AP-619 Assessment out-of-scope journey

stephenrichards:
## What

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@
 # Ignore master key for decrypting credentials and more.
 /config/master.key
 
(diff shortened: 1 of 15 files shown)


reggieb:
In addition for both Services and Predicates, if they are unable to process the data they should fail in the same way - by raising an error. I think we can leave error handling until we have a better idea what errors might occur. I suspect there will be some known errors (like connection to an external service failing) that need to be caught, and the workflow rescheduled to try again later, and unknown errors that should be allowed to bubble to the service to be handled as other system errors.

PR #13, opened 2019-05-31
AP-621 Disposable Capital( liquid, property, vehicles)

stephenrichards:
## Disposable Capital( liquid, property, vehicles)

[Link to story](https://dsdmoj.atlassian.net/browse/AP-621)

This is part of ticket AP-621, which calcualates the assessed values for:

- Liquid Capital
- Property (main dwelling and other properties)
- Vehicles

The `DisposableCapitalAssessment` class is responsible for populating the response section of the particulars object with details of the assessed value of various items, and its interim workings.

To that end, I have added fields into the response schema to show these fields.

## Checklist

### DIFF
--- a/app/services/vehicle_age_calculator.rb
+++ b/app/services/vehicle_age_calculator.rb
@@ -1,6 +1,6 @@
 class VehicleAgeCalculator
   def initialize(purchase_date, calculation_date)
-    @purchase_date = Date.parse(purchase_date)
+    @purchase_date = purchase_date
     @calculation_date = calculation_date
   end
 
--- a/app/services/vehicle_age_calculator.rb
+++ b/app/services/vehicle_age_calculator.rb
@@ -1,0 +1,12 @@
+class VehicleAgeCalculator
+  def initialize(purchase_date, calculation_date)
+    @purchase_date = Date.parse(purchase_date)
+    @calculation_date = calculation_date
+  end
+
+  def in_months
+    months = (@calculation_date.year * 12 + @calculation_date.month) - (@purchase_date.year * 12 + @purchase_date.month)
+    months += 1 if @purchase_date.day < @calculation_date.day
+    months
+  end
+end
--- a/app/services/workflow_predicate/determine_passported.rb
+++ b/app/services/workflow_predicate/determine_passported.rb
@@ -1,6 +1,6 @@
 module WorkflowPredicate
   class DeterminePassported < BaseWorkflowService
-    def result_for
+    def call
       @particulars.request.applicant.receives_qualifying_benefit
     end
   end
--- a/app/services/workflow_predicate/determine_passported.rb
+++ b/app/services/workflow_predicate/determine_passported.rb
@@ -1,7 +1,7 @@
 module WorkflowPredicate
   class DeterminePassported < BaseWorkflowService
     def call
-      @particulars.request.applicant.receives_qualifying_benefit
+      applicant.receives_qualifying_benefit
     end
   end
 end
--- a/app/services/workflow_service/disposable_capital_assessment.rb
+++ b/app/services/workflow_service/disposable_capital_assessment.rb
@@ -19,6 +19,7 @@ module WorkflowService
     end
 
     def calculate_property
+      @remaining_mortgage_allowance = Threshold.value_for(:property_maximum_mortgage_allowance, at: @submission_date)
       calculate_additional_properties
       calculate_main_dwelling(request_data: @particulars.request.applicant_capital.property.main_home,  response_data: @particulars.response.details.capital.property.main_dwelling)
     end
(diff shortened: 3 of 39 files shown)


app/services/vehicle_age_calculator.rb
```diff
@@ -0,0 +1,12 @@
+class VehicleAgeCalculator
+  def initialize(purchase_date, calculation_date)
+    @purchase_date = Date.parse(purchase_date)
+    @calculation_date = calculation_date
+  end
+
+  def in_months
+    months = (@calculation_date.year * 12 + @calculation_date.month) - (@purchase_date.year * 12 + @purchase_date.month)
```
Kalagan: would something like `(@calculation_date - @purchase_date).months` work?

app/services/vehicle_age_calculator.rb
```diff
@@ -0,0 +1,12 @@
+class VehicleAgeCalculator
+  def initialize(purchase_date, calculation_date)
+    @purchase_date = Date.parse(purchase_date)
+    @calculation_date = calculation_date
```
Kalagan: It seems strange to have to parse `purchase_date` but not `calculation_date`.
Ideally, I think it would make sense to parse everything at the beginning of the request and then we only deal with ruby objects/values

app/services/workflow_predicate/determine_passported.rb
```diff
@@ -1,6 +1,6 @@
 module WorkflowPredicate
   class DeterminePassported < BaseWorkflowService
-    def result_for
+    def call
```
Kalagan: I also prefer `call` than `result_for` 👍

app/services/workflow_service/disposable_capital_assessment.rb
```diff
@@ -0,0 +1,21 @@
...
+      VehicleAssessment.new(@particulars).call
+      true
+    end
+
+    private
+
+    def calculate_liquid_capital
+      total_liquid_capital = 0.0
+
+      @particulars.request.applicant_capital.liquid_capital.bank_accounts.each do |acct|
```
Kalagan: I wonder if in `BaseWorkflowService` we could create a method like 
```ruby
def request
  @particulars.request
end
```

It would make this line a bit shorter.

Or even
```ruby
def applicant_capital
  @particulars.request.applicant_capital
end
```

app/services/workflow_service/property_assessment.rb
```diff
@@ -0,0 +1,68 @@
...
+        calculate_individual_property(request_data: additional_property, response_data: response_data, property_type: :additional_property)
+        @particulars.response.details.capital.property.additional_properties << response_data
+      end
+    end
+
+    def calculate_main_dwelling(request_data:, response_data:)
+      calculate_individual_property(request_data: request_data, response_data: response_data, property_type: :main_dwelling)
+    end
+
+    def calculate_individual_property(request_data:, response_data:, property_type:) # rubocop:disable Metrics/AbcSize
```
Kalagan: Ultimately, this could be extracted to its own service

app/services/workflow_service/property_assessment.rb
```diff
@@ -0,0 +1,68 @@
...
+      @remaining_mortgage_allowance = Threshold.value_for(:property_maximum_mortgage_allowance, at: @submission_date)
+      calculate_additional_properties
+      calculate_main_dwelling(request_data: @particulars.request.applicant_capital.property.main_home,
+                              response_data: @particulars.response.details.capital.property.main_dwelling)
+    end
+
+    def calculate_additional_properties
+      @particulars.request.applicant_capital.property.additional_properties.each do |additional_property|
+        response_data = OpenStruct.new(AssessmentParticulars.initial_property_details)
+        calculate_individual_property(request_data: additional_property, response_data: response_data, property_type: :additional_property)
```
Kalagan: I'm not a fan of these `request_data` and `response_data` names...
The `_data` part is not useful and can just be dropped.
Are there better names we can use? Especially for `request_data`:
`property`, `dwelling`, etc... ?

app/services/vehicle_age_calculator.rb
stephenrichards: I don't think so.  `#months` is a method on Fixnum.  Subtracting 1 data from another gives a Rational, representing the number of days.

app/services/vehicle_age_calculator.rb
stephenrichards: I think we agreed to leave this for the time being.

app/services/workflow_service/disposable_capital_assessment.rb
stephenrichards: Agreed.

app/services/workflow_service/property_assessment.rb
stephenrichards: I've replaced them with `property:` and `assessment:`

app/services/workflow_service/property_assessment.rb
stephenrichards: Well we have a property service which handles both main dwelling and an array of additional properties.  The calculations are the same for both, except for a different disregard amount, so I think all in the same place is probably right.

app/services/vehicle_age_calculator.rb
stephenrichards: Ok, have implemented Rob's solution to change them all to dates on the way in.

PR #16, opened 2019-06-07
AP-716 BelowLowerCapitalThresholdPredicate

stephenrichards:
## BelowLowerCapitalThresholdPredicate

This is a simple workflow predicate that compares the lower_capital_threshold and disposable_capital_assessment values in the response structure (which would have been populated by the DisposableCapitalAssessment service) and returns true if the assessment is lower than the threshold.

This predicate will be used in the Standard Workflow

[Link to story](https://dsdmoj.atlassian.net/browse/AP-716)

- Update the workflow to call the the BelowLowerCapitalThresholdPredicate after the DisposableCapitalAssessmentStep
- Wrote the BelowLowerCapitalThresholdPredicate class

## Checklist


app/services/workflow_service/disposable_capital_assessment.rb.orig
```diff
@@ -0,0 +1,50 @@
+module WorkflowService
+  class DisposableCapitalAssessment < BaseWorkflowService
+<<<<<<< HEAD
```
mpw5: Looks like something has gone wrong with a merge here.

mpw5: In fact it looks like this whole file shouldn't have been committed. Should .orig files be added to .gitignore?

PR #17, opened 2019-06-10
AP 640 convert salary values to monthly

reggieb:
[Jira AP-640](https://dsdmoj.atlassian.net/browse/AP-640)

Takes a time series of salary data, and returns the monthly equivalent based on some simple logic:

- if payment weekly or two-weekly: average
- if payment monthly and difference more than 60: average
- if payment monthly and difference less than 60: latest entry

I've assumed the four weekly payments would also be averaged.

The code also includes a time series calculator which includes tools that allows for some analysis of the time periods, and could be used to check patterns match the given time period. This was created when it seemed likely that we would not know the period types, and therefore have to determine them.

### DIFF
--- a/app/services/monthly_salary_calculator.rb
+++ b/app/services/monthly_salary_calculator.rb
@@ -48,7 +48,7 @@ class MonthlySalaryCalculator
   end
 
   def normalize_monthly
-    return time_series_calc.mean if time_series_calc.max_variance > NORMAL_VARIANCE_THRESHOLD
+    return time_series_calc.mean if time_series_calc.max_variance >= NORMAL_VARIANCE_THRESHOLD
 
     time_series_calc.latest_value
   end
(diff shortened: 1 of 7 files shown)


app/services/monthly_salary_calculator.rb
```diff
@@ -0,0 +1,67 @@
...
+  def for_four_weekly
+    weekly_to_monthly(time_series_calc.mean / 4.0)
+  end
+
+  def weekly_to_monthly(weekly)
+    weekly * weeks_in_year / months_in_year
+  end
+
+  def normalize_monthly
+    return time_series_calc.mean if time_series_calc.max_variance > NORMAL_VARIANCE_THRESHOLD
```
Kalagan: Should it be `>=` instead of `>` if we follow the Jira ticket?

PR #18, opened 2019-06-10
AP-544 Add open api doc compatible with Swagger editor

reggieb:
[Jira AP-544](https://dsdmoj.atlassian.net/browse/AP-544)

I've taken the existing JSON Schema, converted them to YAML and then incorporated them into a single OpenAPI document that I think should be compatible with most API documentation systems. 

If you paste `public/schemas/open_api_doc.yaml` into [Swagger Editor](https://swagger.io/tools/swagger-editor/) the output looks like this:
swagger_view_of_assessments

There are a [number of compatible tools](https://apisyouwonthate.com/blog/turning-contracts-into-beautiful-documentation) that should be able to display this documentation. Some of the options include docker solutions (for example [ReDoc](https://github.com/Redocly/redoc#the-docker-way)). I'd suggest mounting that may be simpler than trying to extend the rails app with a gem.

There are tools that will generate JSON Schema from OpenAPI docs, but I struggle to find any that go the other way. Therefore it might make sense to maintain an OpenAPI document as the master description of the API, and use a tool to generate the matching JSON Schema.

Lastly, putting the schema into Swagger editor generated some error and warnings. I think I've fixed the errors, but it would be worth checking the warnings as they suggest there are some incorrectly formatted parts in the current schema.

### DIFF
--- a/public/schemas/open_api_doc.yaml
+++ b/public/schemas/open_api_doc.yaml
@@ -1,0 +1,511 @@
+swagger: "2.0"
+info:
+  description: "This is a [Ministry of Justice](https://www.gov.uk/government/organisations/ministry-of-justice) API for checking financial eligibility."
+  version: "1.0.0"
+  title: "Check Financial Eligibility"
+  termsOfService: "http://swagger.io/terms/"
+  contact:
+    email: "apiteam@swagger.io"
+  license:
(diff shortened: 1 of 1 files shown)


reggieb:
I think our use of apipie has made this redundant - so closing.

PR #22, opened 2019-06-17
AP728 Create Applicant

AbdirizakObsiye2018:
AP728 Create Applicant for Means Calculation Service

[Create Applicant - Means Calculation Service](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&modal=detail&selectedIssue=AP-728)

Created:  

- Applicants model that belongs_to the Assessment model and each assessment has one applicant.
- ApplicantCreationService that validates the applicant params passed in and if successful creates an applicant and returns self.
- ApplicantsController to call the ApplicationCreationService and render a  json success or error response depending on the result.
- Applicant JSON schema
- Tests

## Checklist

### DIFF
--- a/spec/services/applicant_creation_service_spec.rb
+++ b/spec/services/applicant_creation_service_spec.rb
@@ -3,134 +3,128 @@ require 'rails_helper'
 RSpec.describe ApplicantCreationService do
   describe 'POST applicant' do
     let(:assessment) { create :assessment }
-    let(:service) { described_class.new(request_payload) }
+    
+    subject { described_class.call(request_payload) }
 
-    before do
-      # stub requests to get schemas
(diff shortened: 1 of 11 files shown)


spec/services/applicant_creation_service_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+        end
+
+        it 'does not create an applicant' do
+          expect { service.success? }.not_to change { Applicant.count }
+        end
+      end
+    end
+
+    context 'ActiveRecord validation fails' do
+      let(:invalid_payload) do
```
stephenrichards: There is another error condition that needs to be tested (and coded for): if the payload contains an id of an assessment that doesn't exist.

PR #23, opened 2019-06-17
AP-729 refactor of the dependents creation service

stephenrichards:
## Refactor of the DependentsController and DependentCreationService

[Link to story](https://dsdmoj.atlassian.net/browse/AP-729)

Refactored the above classes so as to use the same pattern as Julien used for Assessments creation:
- moved the generation of the response and status code from service to controller
- removed links from response
- made stubbing the response from reading the schema into a rspec helper method

## Checklist

### DIFF
--- a/app/services/dependents_creation_service.rb
+++ b/app/services/dependents_creation_service.rb
@@ -1,10 +1,10 @@
-class DependentCreationService
+class DependentsCreationService
   SCHEMA_PATH = Rails.root.join('public/schemas/dependent.json').to_s
 
   def initialize(raw_post)
     @raw_post =  raw_post
     @payload = JSON.parse(@raw_post, symbolize_names: true)
-    @errors = []
+    @errors = nil
   end
 
   def success?
@@ -12,11 +12,13 @@ class DependentCreationService
   end
 
   def assessment
-    @assessment ||= Assessment.find(@payload[:assessment_id])
+    @assessment ||= Assessment.find_by(id: @payload[:assessment_id])
+    @errors = ['No such assessment id'] if @assessment.nil?
+    @assessment
   end
 
   def errors
-    validator.valid? ? model_errors : validator.errors
+    @errors ||= validator.valid? ? dependent_errors : validator.errors
   end
 
   private
@@ -29,8 +31,12 @@ class DependentCreationService
     @model_errors ||= create_dependents
   end
 
+  def dependent_errors
+    assessment.nil? ? @errors : model_errors
+  end
+
   def create_dependents
-    errors = []
+    ar_errors = []
     ActiveRecord::Base.transaction do
       @payload[:dependents].each do |dependent_params|
         income_params = dependent_params.delete(:income)
@@ -40,10 +46,10 @@ class DependentCreationService
         end
         next if dependent.save
 
-        errors << collect_errors(dependent)
+        ar_errors << collect_errors(dependent)
       end
+      ar_errors.flatten
     end
-    errors.flatten
   end
 
   def collect_errors(dependent)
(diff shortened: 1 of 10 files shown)


app/services/dependents_creation_service.rb
```diff
@@ -0,0 +1,62 @@
+class DependentsCreationService
```
reggieb: I think this class needs refactoring. There are two issues that I think need addressing:

- the error structure is too complicated, 
- and the main action of the object (creating dependents) is only triggered if the `errors` method is called.

So I'd suggest. Create a single error store, and a method to add to it easily.
```ruby
attr_reader :error_store

def add_error(error)
  @errors_store ||= []
  @errors_store << error
end
```
Then everywhere an error can be detected:
```ruby
add_error(something.error) if something.error.present?
```

Then make `create_dependents` public and memoize it so it can be called multiple times - but with only the first one having to do the processing. Then call it at the beginning of every process that relies on it having been run. So:

```ruby
def errors
  create_dependents
  error_store.flatten
end
```
That way other devs don't have to dig through the code to work out where the creation is triggered.

app/services/dependents_creation_service.rb
```diff
@@ -0,0 +1,62 @@
...
+        income_params&.each do |ip|
+          dependent.dependent_income_receipts.new(ip)
+        end
+        next if dependent.save
+
+        ar_errors << collect_errors(dependent)
+      end
+      ar_errors.flatten
+    end
+  end
```
reggieb: I'd split this method in two methods - one method identifies the dependents and then passed the processing of each dependent to a second method:
```ruby
def create_dependents
  @create_dependents ||= ActiveRecord::Base.transaction do
    @payload[:dependents].each do |dependent_params|
      create_dependent(dependent_params)
    end
  end
end

def create_dependent(dependent_params)
  # depenedent creation here
  # with `add_error` used to collect errors as they occur
  ....
end
```

spec/rails_helper.rb
```diff
@@ -75,3 +75,19 @@
...
+def stub_call_to_get_json_schema
+  stub_request(:get, 'http://localhost:3000/schemas/assessment_request.json')
+    .with(
+      headers: {
+        'Accept' => '*/*',
+        'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
+        'User-Agent' => 'Ruby'
+      }
+    )
+    .to_return(status: 200, body: full_schema, headers: {})
```
reggieb: I don't think the headers need to be defined for this stub to work, and their size implies they are important when they are not.
```ruby
stub_request(:get, 'http://localhost:3000/schemas/assessment_request.json')
  .to_return(status: 200, body: full_schema)
```

spec/services/dependents_creation_service_spec.rb
```diff
@@ -1,137 +1,170 @@
...
+    describe 'errors' do
+      it 'returns an error payload' do
+        service.success?
+        expect(service.errors.size).to eq 1
+        expect(service.errors[0]).to eq 'No such assessment id'
+      end
     end
   end
 
+  def payload_with_invalid_id
```
reggieb: Use `let` to define variables in specs:
```
let(:payload_with_invalid_id) do
  {
    ...
```

spec/rails_helper.rb
```diff
@@ -75,3 +75,19 @@
 def open_structify(data)
   JSON.parse(data.to_json, object_class: DatedStruct)
 end
+
+def stub_call_to_get_json_schema
```
reggieb: I think it would be safer and more flexible to pass the schema name into this method:
```ruby
def stub_call_to_schema(schema_name)
  stub_request(:get, "http://localhost:3000/schemas/#{schema_name}.json")
end
```
Then to use it:
```ruby
stub_call_to_schema :assessment_request
```

PR #26, opened 2019-06-19
AP731 Create Properties

AbdirizakObsiye2018:
Create Properties Functionality

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-731)

Created the below:

- Properties model with attributes requested in the story linked above (one attribute in particular, main_home, is a boolean where each assessment has only one property of this type whilst the other properties are additional_properties)

- Properties Creation service to creates properties and return success/error response

- Properties controller that calls the creation service and renders the response result in JSON format with a status code

- Properties Schema

- Tests for properties and updated dependents tests

### DIFF
--- a/app/controllers/properties_controller.rb
+++ b/app/controllers/properties_controller.rb
@@ -1,0 +1,6 @@
+class PropertiesController < ApplicationController
+  def create
+    result = PropertiesCreationService.call(request.raw_post)
+    render json: result.to_h, status: result.success ? 200 : 422
+  end
+end
--- a/app/models/property.rb
+++ b/app/models/property.rb
@@ -1,5 +1,5 @@
 class Property < ApplicationRecord
-  # belongs_to :asssessment
+  belongs_to :assessment
 
   validates :value, numericality: { greater_than: 0 }
   validates :outstanding_mortgage, numericality: { greater_than_or_equal_to: 0 }
--- a/app/models/property.rb
+++ b/app/models/property.rb
@@ -1,0 +1,8 @@
+class Property < ApplicationRecord
+  # belongs_to :asssessment
+
+  validates :value, numericality: { greater_than: 0 }
+  validates :outstanding_mortgage, numericality: { greater_than_or_equal_to: 0 }
+  validates :percentage_owned, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 100 }
+  validates :main_home, :shared_with_housing_assoc, inclusion: { in: [true, false] }
+end
--- a/app/services/properties_creation_service.rb
+++ b/app/services/properties_creation_service.rb
@@ -24,15 +24,15 @@ class PropertiesCreationService
   private
 
   def json_valid?
-    validator = JsonSchemaValidator.new(raw_post, SCHEMA_PATH)
+    validator = JsonSchemaValidator.new(@raw_post, SCHEMA_PATH)
     return true if validator.valid?
     @errors = validator.errors
     false
   end
 
   def assessment_exists?
-    assessment = Assessment.find_by(id: @payload[:assessment_id])
-    if assessment.nil?
+    @assessment = Assessment.find_by(id: @payload[:assessment_id])
+    if @assessment.nil?
       @errors << 'No such assessment id'
       return false
     end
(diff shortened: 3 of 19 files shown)


app/services/properties_creation_service.rb
```diff
@@ -0,0 +1,86 @@
...
+  def initialize(raw_post)
+    @raw_post = raw_post
+    @payload = JSON.parse(@raw_post, symbolize_names: true)
+    @errors = []
+  end
+
+  def call
+    if json_valid? && assessment_exists?
+      return success_response if create_properties
+    end
```
Kalagan: You might as well write
`return success_response if json_valid? && assessment_exists? && create_properties`
rather than an `if` inside another

app/controllers/properties_controller.rb
```diff
@@ -0,0 +1,6 @@
+class PropertiesController < ApplicationController
+  def create
+    result = PropertiesCreationService.call(request.raw_post)
+    render json: result.to_h, status: result.success ? 200 : 422
```
Kalagan: Like Rob suggested on slack you could try to implement the `to_json` method in the service and then writing just `json: result` here.

app/models/property.rb
```diff
@@ -0,0 +1,8 @@
+class Property < ApplicationRecord
+  belongs_to :assessment
+
+  validates :value, numericality: { greater_than: 0 }
+  validates :outstanding_mortgage, numericality: { greater_than_or_equal_to: 0 }
+  validates :percentage_owned, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 100 }
+  validates :main_home, :shared_with_housing_assoc, inclusion: { in: [true, false] }
```
Kalagan: Aren't these validations duplicates of what is already checked by the JSON schema?

spec/requests/properties_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        expect(response).to have_http_status(:unprocessable_entity)
+      end
+
+      it 'returns the expected error response payload' do
+        expect(response.body).to eq error_response.to_h.to_json
+      end
+    end
+  end
+
+  def success_response
```
Kalagan: you should use the `let` keyword here rather than a method. To follow RSpec's convention

spec/services/properties_creation_service_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+
+RSpec.describe PropertiesCreationService do
+  before { stub_call_to_json_schema }
+
+  let(:assessment) { create :assessment }
+
+  shared_examples 'error response' do
+    it 'does not create any property records' do
+      expect {
+        described_class.call(payload)
```
Kalagan: To follow RSpec's convention, you can define `subject` at the top (just below `let(:assessment)`) and then always use `subject`

```ruby
subject { described_class.call(payload) }
...
...
expect { subject }.not_to change { Property.count }
```

spec/services/properties_creation_service_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+    payload = valid_payload_hash.dup
+    payload[:extra_root_attr] = 1
+    payload.delete(:assessment_id)
+    payload[:properties][:main_home][:extra_main_home_attr] = 3
+    payload[:properties][:additional_properties].first[:extra_main_home_attr] = 3
+    payload[:properties][:main_home].delete(:percentage_owned)
+    payload.to_json
+  end
+
+  def invalid_assessment_payload
```
Kalagan: you can use `let` for all these methods

app/controllers/properties_controller.rb
stephenrichards: @Julien - We're going round in circles here.  My first implementation was for the service to generate the json response and https status, and you refactored it so that the controller did this.  I think that is probably better - the service is responsible for creating the records and generating the data that's going into the response, the controller is responsible for formatting it to JSON, XML, HTML or whatever we decide.  I know we're only going to use JSON here, but the principle holds good.

julien: are we?

PR #30, opened 2019-07-02
AP-734 create outgoings

reggieb:
[Jira AP-734](https://dsdmoj.atlassian.net/browse/AP-734)

This PR creates an outgoings endpoint.

It also uses [apipie](https://github.com/Apipie/apipie-rails) to document and validate the JSON requests/responses.

The PDF below shows an HTML version of the resultant api docs for the new create action (the real thing is a little prettier). This is available at http://localhost:3000/apidocs/1.0/outgoings/create.html

[API documentation.htm.pdf](https://github.com/ministryofjustice/check-financial-eligibility/files/3350730/API.documentation.htm.pdf)


app/controllers/outgoings_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  private
+
+  def outgoing_creation_service
+    @outgoing_creation_service ||= OutgoingsCreationService.call(input[:outgoings])
+  end
+
+  def input
+    @input ||= JSON.parse(request.raw_post).symbolize_keys
```
Kalagan: would
```ruby
JSON.parse(request.raw_post, symbolize_names: true)
```
work instead of 
```ruby
JSON.parse(request.raw_post).symbolize_keys
```
?

app/models/outgoing.rb
```diff
@@ -0,0 +1,9 @@
+class Outgoing < ApplicationRecord
+  enum type_of_outgoing: {
+    rent: 'rent',
+    mortgage: 'mortgage',
+    child_care: 'child_care',
+    maintenance: 'maintenance',
+    legal_aid: 'legal_aid'
```
Kalagan: no `.freeze`? 😃

app/controllers/outgoings_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      render json: outgoing_creation_service
+    else
+      render_unprocessable(outgoing_creation_service.error_message)
+    end
+  end
+
+  private
+
+  def outgoing_creation_service
+    @outgoing_creation_service ||= OutgoingsCreationService.call(input[:outgoings])
```
Kalagan: Don't you need to also pass `input[:assessment_id]` in addition to `input[:outgoings]`?

app/services/outgoings_creation_service.rb
```diff
@@ -0,0 +1,33 @@
...
+      success: success?,
+      outgoings: (outgoings if success?),
+      errors: error_message
+    }
+  end
+
+  def outgoings
+    @outgoings ||= Outgoing.transaction do
+      params.map { |attrs| Outgoing.create(attrs) }
+    end
```
Kalagan: These outgoings don't belong to any assessment?
I think you need to get the `assessment` and then you should be able to do `assessment.outgoings.create!(params)`

app/controllers/outgoings_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    param :payment_date, Date, date_option: :today_or_older, required: true, desc: 'The date the outgoing was paid'
+    param :amount, :currency, required: true, desc: 'The monetary amount paid for the outgoing'
+  end
+
+  returns code: :ok, desc: 'Successful response' do
+    property :objects, array_of: Outgoing
+    property :success, ['true'], desc: 'Success flag shows true'
+  end
+  returns code: :unprocessable_entity do
+    property :error, String, desc: 'Description of why object invalid'
```
Kalagan: I think errors should be an array like we decided in our slack conversation with the other devs

app/controllers/outgoings_controller.rb
```diff
@@ -0,0 +1,38 @@
+class OutgoingsController < ApplicationController
+  # api! works but the links in resulting docs are broken. See https://github.com/Apipie/apipie-rails/issues/559
+  api :POST, 'assessments/:assessment_id/outgoings', 'Create outgoings'
+  formats ['json']
+  param :assessment_id, :uuid, required: true
+  param :outgoings, Array, desc: 'Collection of outgoings', required: true do
+    param :type_of_outgoing, Outgoing.type_of_outgoings.values, required: true, desc: 'The type of product or service the outgoing paid for'
+    param :payment_date, Date, date_option: :today_or_older, required: true, desc: 'The date the outgoing was paid'
```
Kalagan: i wonder if we should use `:date` instead of `Date` for consistency because we use `:uuid` and `:currency`

db/migrate/20190624151225_create_outgoings.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOutgoings < ActiveRecord::Migration[5.2]
+  def change
+    create_table :outgoings, id: :uuid do |t|
+      t.string :type_of_outgoing
+      t.date :payment_date
+      t.decimal :amount
+
```
Kalagan:
```ruby
t.belongs_to :assessment, foreign_key: true, null: false, type: :uuid
```
?

db/migrate/20190624151225_create_outgoings.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOutgoings < ActiveRecord::Migration[5.2]
+  def change
+    create_table :outgoings, id: :uuid do |t|
+      t.string :type_of_outgoing
```
Kalagan: `:outgoing_type` ?

spec/models/outgoing_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Outgoing, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
Kalagan: We can remove this file if we don't need it.

app/services/outgoings_creation_service.rb
reggieb: Good point - thank you.

db/migrate/20190624151225_create_outgoings.rb
reggieb: Yes - that's better. But [the ticket](https://dsdmoj.atlassian.net/browse/AP-734) has "type_of_outgoing" in the JSON input and I think the model name should match that. I'll check with Stephen and see if we can change it.

app/controllers/outgoings_controller.rb
reggieb: `Date` matches the usage in other places. For example line 6 `param :outgoings, Array, ....`

If you look at the apipie documentation, it appears that the default is to pass in the object class if you can, and then use a :symbol description when that's not possible.

Kalagan: Alright that makes sense

db/migrate/20190624151225_create_outgoings.rb
reggieb: Stephen has confirmed the name can change.

reggieb: I've change the name in the next migration

Kalagan: An alternative would be to:
- doing a rollback
- changing the name of this enum and adding the assessment reference
- running the migration again

I think that's acceptable because this is a branch and not deployed to staging or prod. Even if it's deployed to UAT, I would delete the deployment before redeploying

Kalagan: It would avoid adding an extra migration to correct a previous migration which is not merged.

spec/requests/outgoings_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    subject { post assessment_outgoings_path(assessment), params: params.to_json, headers: headers }
+
+    it 'returns http success', :show_in_doc do
+      subject
+      expect(response).to have_http_status(:success)
+    end
+
+    it 'creates outgoings' do
+      expect { subject }.to change { Outgoing.count }.by(2)
```
Kalagan: `expect { subject }.to change { assessment.outgoings.count }.by(2)` would be better

app/controllers/outgoings_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    param :amount, :currency, required: true, desc: 'The monetary amount paid for the outgoing'
+  end
+
+  returns code: :ok, desc: 'Successful response' do
+    property :objects, array_of: Outgoing
+    property :success, ['true'], desc: 'Success flag shows true'
+  end
+  returns code: :unprocessable_entity do
+    property :error, array_of: String, desc: 'Description of why object invalid'
```
Kalagan: isn't it `errors` instead of `error` ?

spec/requests/outgoings_spec.rb
reggieb: Yup - did that in the service spec. Missed it here. Updated.

db/migrate/20190624151225_create_outgoings.rb
reggieb: I don't like modifying migrations across commits, but you're right - it's the merge that's the more important cutoff.

Kalagan: You can modify it within the same commit. Create a new commit and then move it around and merge it with the previous commit with an interactive rebase.

reggieb: I've squashed the whole PR.

Kalagan: That works too 😄

PR #34, opened 2019-07-09
AP-769 Convert Incomes controller to use Apipie

reggieb:
[Jira Ticket AP-769](https://dsdmoj.atlassian.net/browse/AP-769)

Converts the incomes controller to use apipie rather than the JSON schema


spec/requests/incomes_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    it 'sets success flag to false' do
+      subject
+      expect(parsed_response[:success]).to be true
+    end
+
+    it 'returns blank errors' do
+      subject
+      expect(parsed_response[:errors]).to be_empty
+    end
+
```
stephenrichards: I think we should have a spec which checks that the response contains the created benefit receipts and wage slips.

PR #36, opened 2019-07-17
AP-783 Adds PaymentPeriodAnalyser to determine frequency of payment

reggieb:
[Jira AP-783](https://dsdmoj.atlassian.net/browse/AP-783)

Generates sets of time series data of various patterns, and passes them to a new service `PaymentPeriodAnalyser` to determine if it can identify the correct period of payment.

The solution works, but struggles to distinguish between monthly and four-weekly series where there is a lot of variance (more than a couple of days) and the sequence of days is negatively sloped (that is, each day is earlier in the month). That is, _[13-May-19, 14-Jun-19, 15-Jul-19]_ is unlikely to be four weekly because each day is higher through the sequence (positively sloped). But _[16-May-19, 14-Jun-19, 12-Jul-19]_ could be either four weekly or monthly with a two day variance.

### DIFF
--- a/app/services/payment_period_analyser.rb
+++ b/app/services/payment_period_analyser.rb
@@ -1,0 +1,102 @@
+# rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+class PaymentPeriodAnalyser
+  attr_reader :data
+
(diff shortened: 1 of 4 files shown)


reggieb:
Merging code so that it is available for reworking and productionising in AP-641

PR #37, opened 2019-07-18
Ap 768 convert dependants controller to APIPIE

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-768)

Converts the dependants controller to use apipie rather than the JSON schema, I had to add ```active_model_serializer``` gem to be able to display the attributes [```:date_of_payment```, ```amount```] nested under ```dependant_income_receipts```.

## Checklist


spec/requests/dependants_spec.rb
```diff
@@ -4,59 +4,125 @@
...
+      before { post assessment_dependants_path(assessment), params: request_payload.to_json, headers: headers }
+
+      it 'returns http unprocessable entity' do
+        expect(response).to have_http_status(:unprocessable_entity)
+      end
+
+      it 'returns error payload' do
+        expect(parsed_response[:success]).to eq(false)
+        expect(parsed_response[:errors]).to_not be_empty
       end
```
stephenrichards: I don't think testing that there are errors is enough - we should test for the error message we are expecting, otherwise it may be failing for some reason we don't know about

spec/requests/dependants_spec.rb
```diff
@@ -4,59 +4,125 @@
...
+
+      it 'returns an error and is shown in apidocs', :show_in_doc do
         expect(response).to have_http_status(:unprocessable_entity)
       end
 
       it 'returns error payload' do
         expect(parsed_response[:success]).to eq(false)
-        expect(parsed_response[:objects]).to eq(nil)
-        expect(parsed_response[:errors]).to match_array(%w[error_1 error_2])
+        expect(parsed_response[:errors]).to_not be_empty
```
stephenrichards: Same re testing for the expected error message

spec/requests/dependants_spec.rb
```diff
@@ -4,59 +4,125 @@
...
       end
 
+      it 'errors and is shown in apidocs', :show_in_doc do
+        expect(response).to have_http_status(422)
+      end
+
       it 'returns error payload' do
-        expect(parsed_response[:errors]).not_to be_empty
+        expect(parsed_response[:success]).to eq(false)
+        expect(parsed_response[:errors]).to_not be_empty
```
stephenrichards: same re testing for the expected errror message

PR #43, opened 2019-07-24
AP-641 identify frequency of payments

reggieb:
[Jira AP-641](https://dsdmoj.atlassian.net/browse/AP-641)

This adds a `payment_pattern` class method to `BenefitReceipt` and `WageSlip`, together with a `time_series` scope option. With this in place, the payment patterns for an assessment can be determined via:

- `assessment.benefit_receipts.payment_pattern`
- `assessment.wage_slips.payment_pattern`

`payment_pattern` returns one of `:monthly`, `:weekly`, `:two_weekly`, `:four_weekly`, `:unknown` depending on the sequence of payment dates.

Also the slope calculation has been split out from `PaymentPeriodAnalyser` and a system added that allows better analysis of data sequences that straddle the month end. Also constants have been added to `PaymentPeriodAnalyser` to aid adjustment of the values at a later date.

### DIFF
--- a/app/models/benefit_receipt.rb
+++ b/app/models/benefit_receipt.rb
@@ -6,6 +6,15 @@ class BenefitReceipt < ApplicationRecord
 
   validate :payment_date_cannot_be_in_future
 
+  scope :time_series, -> { pluck(:payment_date, :amount).to_h.transform_keys(&:to_time) }
+
+  scope :payment_pattern, -> do
+    return :no_data unless time_series.present?
+
+    analyser = PaymentPeriodAnalyser.new(time_series)
+    analyser.period_pattern
+  end
+
   def payment_date_cannot_be_in_future
     errors.add(:base, 'Benefit payment date cannot be in the future') if payment_date > Date.today
   end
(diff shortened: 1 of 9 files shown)


app/models/benefit_receipt.rb
```diff
@@ -1,5 +1,8 @@
 class BenefitReceipt < ApplicationRecord
   extend EnumHash
+  extend PaymentPatternConcern
+  define_payment_pattern
```
reggieb: Not too sure about this method name. `has_payment_pattern` seemed a better fit, but rubocop complained.

app/services/date_slope.rb
```diff
@@ -0,0 +1,73 @@
...
+  def days
+    @days ||= DescriptiveStatistics::Stats.new(dates.map(&:day))
+  end
+
+  def normalized_days
+    @normalized_days ||= DescriptiveStatistics::Stats.new(days_at_beginning_bump_to_previous_month)
+  end
+
+  def best_match_days
+    return days if days.range < 10
```
Kalagan: what's the meaning of `10` ?

reggieb: Good point - I've moved this (and the other 10) out to constants with comments

PR #46, opened 2019-07-26
AP-786 Amend NonLiquidCapitalAssessment service

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-785)

Refactor the `NonLiquidCapitalAssessment` service to use data from the database and not the JSON struct.

- Accept `assessment_id` rather than JSON struct in `initializer`
- Use the `Assessment` object to retrieve `non_liquid_assets` data

This PR doesn't change the DisposableCapitalAssessment service which calls NonLiquidCapitalAssessment. That will be handled in a separate story (AP-784).

## Checklist

### DIFF
--- a/app/models/dated_struct.rb
+++ b/app/models/dated_struct.rb
@@ -13,9 +13,12 @@ class DatedStruct < OpenStruct
     end
   end
 
+  # TODO: remove this method (or even the whole class) once all reference to it has been removed from other services
+  # :nocov:
   def []=(name, value)
     modifiable?[new_ostruct_member!(name)] = value_or_time(value)
   end
+  # :nocov:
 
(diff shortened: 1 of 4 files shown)


app/models/dated_struct.rb
```diff
@@ -13,9 +13,11 @@ def initialize(hash = nil, options = {})
     end
   end
 
+  # :nocov:
   def []=(name, value)
     modifiable?[new_ostruct_member!(name)] = value_or_time(value)
   end
+  # :nocov:
```
stephenrichards: let's put a # TODO to remind us to remove this method once all reference to it has been removed from other services.

PR #49, opened 2019-07-31
Ap 789 vehicle assessment service db

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-759)

Refactor the `VehicleAssessment` service to use data from the database and not the JSON struct.

* Accept an `Assessment` instance rather than JSON struct in `initializer`
* Use the `Assessment` object to retrieve `Vehicle` data
* Fix bug in `assessment_factory` by switching to `Faker` to generate IP addresses

## Checklist

### DIFF
--- a/app/services/base_workflow_service.rb
+++ b/app/services/base_workflow_service.rb
@@ -28,4 +28,8 @@ class BaseWorkflowService
   def additional_properties
     @additional_properties ||= @assessment.properties.where(main_home: false)
   end
+
+  def vehicles
+    @vehicles ||= @assessment.vehicles
+  end
 end
--- a/app/services/workflow_service/vehicle_assessment.rb
+++ b/app/services/workflow_service/vehicle_assessment.rb
@@ -1,23 +1,21 @@
 module WorkflowService
-  class VehicleAssessment
-    def initialize(vehicles, submission_date)
-      @vehicles = vehicles
-      @submission_date = submission_date
-      @response = []
-    end
-
+  class VehicleAssessment < BaseWorkflowService
     def call
-      @vehicles.each { |v| assess(v) }
+      vehicles.each { |v| assess(v) }
       @response
     end
 
     private
 
+    def response
+      @response ||= []
+    end
+
     def assess(vehicle)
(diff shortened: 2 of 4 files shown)


app/services/base_workflow_service.rb
```diff
@@ -28,4 +28,8 @@ def main_home
   def additional_properties
     @additional_properties ||= @assessment.properties.where(main_home: false)
   end
+
+  def vehicles
+    @vehicles || @assessment.vehicles
```
Kalagan: Did you mean to write `||=`?
Although this memoization and most of the others in this file are not needed because ActiveRecord already caches the result.
If you call `assessment.vehicles` twice, only the first query is made to the DB and the result is cached.
That's why we often use `.reload` in the specs to force re-fetching the records.

So then the method becomes:
```ruby
def vehicles
  @assessment.vehicles
end
```

And then this can be replaced by adding an attribute and a delegate:
```ruby
class BaseWorkflowService
  attr_reader :assessment

  delegate :vehicles, to: :assessment

  def initialize(assessment)
    @assessment = assessment
```
Which is nicer in my opinion.
If we go that route, then we might as well do it for the other methods!

```ruby
class BaseWorkflowService
  attr_reader :assessment

  delegate :applicant, :submission_date, :bank_accounts, :non_liquid_assets, :result, :main_home, :additional_properties, :vehicles, to: :assessment

  def initialize(assessment)
    @assessment = assessment
    @calculation_period = CalculationPeriod.new(submission_date)
  end
end
```
```ruby
class Assessment < ApplicationRecord
  ...

  def main_home
    @main_home ||= properties.find_by(main_home: true)
  end

  def additional_properties
    @additional_properties ||= properties.where(main_home: false)
  end
```

But this is not at all the subject of this PR so please don't worry about it 😂 
We can refactor later.
You can just add the missing `=` or remove the `@vehicles || `

mpw5: Thanks. Good suggestions. I've add the `=` for now to keep it consistent with the other methods. I'll refactor it later.

PR #55, opened 2019-08-01
Fix failing capital spec

mpw5:
## What
The `capitals_creation_service_spec` randomly fails. This is due to the use of `Faker::Bank.name` to create bank account names. There is only a small pool of bank names to select from so occasionally both accounts are created with the same name, leading to failures when `find_by!(name: bank_account[:name])` is executed.

I've fixed this by adding `Faker::Bank.account_number` to the name created by FactoryBot, which should hopefully result in unique names always being created.

## Checklist

### DIFF
--- a/spec/factories/bank_account_factory.rb
+++ b/spec/factories/bank_account_factory.rb
@@ -1,7 +1,7 @@
 FactoryBot.define do
   factory :bank_account do
     assessment
-    name { Faker::Bank.name }
+    name { "#{Faker::Bank.name} #{Faker::Bank.unique.account_number}" }
     lowest_balance { Faker::Number.decimal(4, 2).to_f }
   end
 end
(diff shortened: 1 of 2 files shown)


Kalagan:
An another way to make it unique is to use FactoryBot's sequence:
```
sequence(:name) { |n| "#{Faker::Bank.name} #{n}" }
```
`n` is incremented each time

mpw5:
> You could use `"#{Faker::Bank.name} #{Faker::Bank.unique.account_number}"` to be sure it's unique and avoid the 22 limit.
> `account_number` seems to have much more unique values. (https://github.com/stympy/faker/blob/8eff383dfe8a05f9f896f1bad13b2c66733fdb6a/lib/faker/default/bank.rb#L8)

I wasn't aware of the `unique` attribute for Faker. I've added that to the spec. That should give us more than enough names for our tests.

PR #64, opened 2019-08-06
Align dependant model with real data

Kalagan:
## What

Add missing dependant's attributes and replace the array of incomes with a simple `monthly_income` attribute.
As agreed with Stephen and Dave

## Checklist


app/models/dependant.rb
```diff
@@ -1,7 +1,10 @@
 class Dependant < ApplicationRecord
   belongs_to :assessment
-  has_many :dependant_income_receipts
-  accepts_nested_attributes_for :dependant_income_receipts
+
+  enum relationship: {
+    child_relative: 'child_relative'.freeze,
+    adult_relative: 'adult_relative'.freeze
+  }
```
reggieb: You could use `enum_hash_for` here:
```ruby
enum relationship: enum_hash_for(:child_relative, :adult_relative)
```

Kalagan: Good point. I saw you added this the other day. Thanks for the reminder

reggieb: It might be worth updating `enum_hash_for` so that it freezes the created strings.

PR #69, opened 2019-08-20
AP-885 Refactor to hold results and interim calculations in DB

stephenrichards:
## Refactor to hold results and interim calculations in database

[Link to story](https://dsdmoj.atlassian.net/browse/AP-885)

- Major refactor to make a new `CapitalSummary` model which will hold summary results from the capital assessment
- Created `capital_items` table to hold `LiquidCapitalItem` and `NonLiquidCapitalItem` records which all belong to the `CapitalSummary`
- Replaced the `NonLiquidCapitalAsset` model with the `NonLiquidCapitalItem`model
- Replaced the `BankAccount` model with the `LiquidCapitalItem` model
- Added interim result fields to the `Property`, `Vehicle` models
- Deleted the `Result` model - results and interim results will be stored on the models to which they relate, and summarised in the `CapitalSummary`
- Removed code relating to Income - this was written using the JSON structure and is not required for launch (passported only)
- Amended the Vehicle, Property, Liquid and NonLiquid Capital Assessment services to write interim results to the new fields on the respective record, rather than a catch-all Result model
- Removed code in `VehiclesAssessment` service that applied a discount to older vehicles (no longer required), and associated config data from the Thresholds configuration files
- Moved most of the assessment logic from the service to the Property, Vehicle models, so that the services just deal with iterating through the collection of records rather than performing the calculations.
- Removed `LegacyBaseWorkflow` service (used only during the transition from JSON payload rather than database)
- Added more detailed API documentation  for the Capitals, Applicant and Assessment endpoints
- changed Capitals payload to be just two arrays - one for liquid capital and one for non-liquid capital
- Re-instatated the specs that had been marked as pending while the refactor took place

## What has *not* been done

- Standardised a response layout, in particular how to render associated objects
- The final Disposable Capital Assessment that will work out whether the applicant is eligible, not eligible, or eligible with a contribution (and work out what that contribution is)

## Checklist


app/controllers/assessments_controller.rb
```diff
@@ -1,7 +1,31 @@
...
+      This endpoint should be called first to create an Assessment. Then using the
+      assessment_id returned by this call, other resources should be called to build
+      up the complete set of data relating to the assessment:
+
+        POST /assessments/:assessment_id/applicant      # adds data about the applicant
+        POST /assessments/:assessment_id/capitals       # adds data about liquid assets (i.e. bank accounts) and
+                                                        # non-liquid assets (valuable items, trusts, etc)
+        POST /assessments/:assessment_id/properties     # adds data about properties owned by the applicant
+        POST /assessments/:assessment_id/vehicles       # adds data about vehicles owned by the applicant
+        POST /assessments/:assessment_id/dependants     # adds data about any dependents the applicant may have
```
mpw5: `dependents` in the comment should be `dependants`

app/controllers/assessments_controller.rb
```diff
@@ -1,7 +1,31 @@
...
+        POST /assessments/:assessment_id/incomes        # adds data about any income the applicant may have
+
+      Once all the above calls have been made to build up a complete picture of the applicant's assets and income
+      the followin call should be made to perform the assessment and get the result:
```
mpw5: Missing `g` at the end of `followin`

app/controllers/assessments_controller.rb
```diff
@@ -1,7 +1,31 @@
...
+
+        GET /assessment/:assessment_id
+
+    END_OF_TEXT
+  end
   api :POST, 'asssessments', 'Create asssessment'
```
mpw5: It wasn't changed in this PR, but both `asssesments` on this line contain an extra `s`

app/services/workflow_service/property_assessment.rb
```diff
@@ -2,60 +2,70 @@ module WorkflowService
...
+      Threshold.value_for(:property_notional_sale_costs_percentage, at: @submission_date) / 100.0
     end
 
-    def assessed_capital_value(assessment:)
-      [0, (assessment.net_equity_value - assessment.property_disregard).round(2)].max
+    def percentage_owned(property)
+      property.percentage_owned / 100.0
+    end
+
+    def asssessed_equity(property)
```
mpw5: Three `s`s in `asssessed`

spec/services/capitals_creation_service_spec.rb
```diff
@@ -15,58 +21,104 @@
...
-        expect(subject.capital[:non_liquid_assets].first.assessment.id).to eq(assessment.id)
+    context 'non_liquid_capital_items_only' do
+      let(:non_liquid_assets) { non_liquid_assets_hash }
+
+      before { subject }
+
+      it 'creates non liquid capital items' do
+        expect(capital_summary.non_liquid_capital_items.size).to eq 1
+        expect(capital_summary.non_liquid_capital_items.first.description).to eq item_1
+        expect(capital_summary.non_liquid_capital_items.first.value).to eq value_1
```
mpw5: do we also need a `it 'does not create liquid capital items'` test here, similar to the one above where we test that creating liquid capital items does not create non-liquid items?

spec/services/vehicle_creation_service_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe VehicleCreationService do
   let(:assessment) { create :assessment }
+  let(:capital_summary) { assessment.capital_summary }
```
mpw5: `capital_summary` is declared here but doesn't seem to be used?

spec/services/workflow_service/pensioner_capital_disregard_spec.rb
```diff
@@ -3,73 +3,36 @@
 module WorkflowService
   RSpec.describe PensionerCapitalDisregard do
     let(:service) { described_class.new(assessment) }
-    let(:request_hash) { AssessmentRequestFixture.ruby_hash }
     let(:assessment) { create :assessment, applicant: applicant }
+    let(:capital_summary) { assessment.capital_summary }
```
mpw5: capital_summary is declared here but doesn't seem to be used?

app/models/capital_summary.rb
```diff
@@ -0,0 +1,25 @@
...
+  enum(
+    capital_assessment_result: {
+      pending: 'pending'.freeze,
+      summarised: 'summarised'.freeze,
+      eligible: 'eligible'.freeze,
+      not_eligible: 'not eligible'.freeze,
+      contribution_required: 'contribution required'.freeze
+    },
+    _prefix: false
+  )
```
reggieb: I'd suggest - to be consistent with other models:
```ruby
enum capital_assessment_result: enum_hash_for(:pending, :summarised, :eligible, :not_eligible, :contribution_required), _prefix: false
```

app/models/capital_summary.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def main_home
+    properties.where(main_home: true).first
```
reggieb: I'd use `find_by` here:
```ruby
properties.find_by(main_home: true)
```

app/services/base_workflow_service.rb
```diff
@@ -5,31 +5,31 @@ def initialize(assessment)
...
-  def bank_accounts
-    @bank_accounts ||= @assessment.bank_accounts
-  end
+  attr_reader :assessment
 
-  def non_liquid_assets
-    @non_liquid_assets ||= @assessment.non_liquid_assets
+  def applicant
+    @applicant ||= @assessment.applicant
   end
```
reggieb: Both `applicant` and `capital_summary` can be delegated from `assessment`. Then `main_home`, `additional_properties`, `liquid_capital_items` and `vehicles` can be delegated from `capital_summary`.

app/services/capitals_creation_service.rb
```diff
@@ -12,29 +12,36 @@ def call
     self
   end
 
+  def capital_summary
+    assessment.capital_summary
+  end
```
reggieb: I'd delegate this.

app/services/properties_creation_service.rb
```diff
@@ -38,10 +38,14 @@ def new_additional_properties
...
+    @properties << capital_summary.properties.create!(attrs)
   end
 
   def assessment
     @assessment ||= Assessment.find_by(id: assessment_id) || (raise CreationError, ['No such assessment id'])
   end
+
+  def capital_summary
+    assessment.capital_summary
+  end
```
reggieb: Delegate again.

app/services/workflow_service/disposable_capital_assessment.rb
```diff
@@ -1,42 +1,56 @@
...
-      VehicleAssessment.new(applicant_capital.vehicles, @submission_date).call
+      VehicleAssessment.new(assessment).call
+    end
+
+    def sum_capital_totals
+      capital_summary.total_capital = capital_summary.total_liquid +
+                                      capital_summary.total_non_liquid +
+                                      capital_summary.total_vehicle +
+                                      capital_summary.total_property
+    end
```
reggieb: I think this logic belongs within the `CapitalSummary`.

app/services/workflow_service/property_assessment.rb
```diff
@@ -2,60 +2,70 @@ module WorkflowService
   class PropertyAssessment < BaseWorkflowService
```
reggieb: Most of the methods in this service pass in a property, and a single instance of the service does all the calculations for all the properties. This strongly suggests that the structure of this object can be improved.

Firstly, I would refactor the class so that a class method `call` receives `capital_summary`. Then have that method iterate through all the properties and initiate an instance of the service to handle just that property. So:

```ruby
def self.call(capital_summary)
  capital_summary.properties.sum { |property| new(property, capital_summary).assessed_equity }
end
```

Second, each property should knows whether it is the main home. So you don't have to split out separate methods do handle the processing of each property type, but rather just check the `main_home` attribute where it needs to be checked.

In this way this class can be simplified a lot.

app/services/workflow_service/vehicle_assessment.rb
```diff
@@ -2,7 +2,7 @@ module WorkflowService
   class VehicleAssessment < BaseWorkflowService
     def call
       vehicles.each { |v| assess(v) }
-      @response
+      vehicles.sum(&:assessed_value)
```
reggieb: Again - pass the collection of vehicles in at the class level and pass each one to an instance of the class.

app/controllers/capitals_controller.rb
```diff
@@ -1,12 +1,28 @@
...
   api :POST, 'assessments/:assessment_id/capitals', 'Create capital declarations'
   formats ['json']
   param :assessment_id, :uuid, required: true
-  param :liquid_capital, Hash, desc: 'Defines liquid capital assets' do
-    param :bank_accounts, Array, required: true, desc: 'Collection of bank accounts' do
-      param :name, String, required: true, desc: 'An identifying name for this bank account'
-      param :lowest_balance, :currency, required: true, desc: 'The lowest balance on this bank account during the calculation period'
-    end
+  param :liquid_capital, Array, desc: 'Collection of bank accounts' do
+    param :description, String, required: true, desc: 'An identifying name for this bank account'
```
Kalagan: I like the fact that you removed the extra hierarchy. But wouldn't `bank_accounts` more sense than `liquid_capital` here?
Could it be something other than a bank account?

db/schema.rb
```diff
@@ -56,6 +48,35 @@
...
+    t.decimal "total_mortgage_allowance", default: "0.0", null: false
+    t.decimal "total_capital", default: "0.0", null: false
+    t.decimal "pensioner_capital_disregard", default: "0.0", null: false
+    t.decimal "assessed_capital", default: "0.0", null: false
+    t.decimal "capital_contribution", default: "0.0", null: false
+    t.decimal "lower_threshold", default: "0.0", null: false
+    t.decimal "upper_threshold", default: "0.0", null: false
+    t.string "capital_assessment_result", default: "pending", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
```
reggieb: Use `t.timestamps` rather than manually defining `created_at` and `updated_at`

db/schema.rb
```diff
@@ -56,6 +48,35 @@
...
+    t.index ["capital_summary_id"], name: "index_capital_items_on_capital_summary_id"
+  end
+
+  create_table "capital_summaries", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "assessment_id"
+    t.decimal "total_liquid", default: "0.0", null: false
+    t.decimal "total_non_liquid", default: "0.0", null: false
+    t.decimal "total_vehicle", default: "0.0", null: false
+    t.decimal "total_property", default: "0.0", null: false
+    t.decimal "total_mortgage_allowance", default: "0.0", null: false
```
reggieb: Do we really need all these defaults and null declarations?

spec/factories/capital_item_factory.rb
```diff
@@ -0,0 +1,17 @@
+FactoryBot.define do
+  factory :non_liquid_capital_item do
+    capital_summary
+    description { Faker::Lorem.unique.sentence }
+    value { Faker::Number.decimal(l_digits: 4, r_digits: 2).to_f }
```
reggieb: Should these value attributes use `to_d`, rather than `to_f`

spec/factories/capital_summary_factory.rb
```diff
@@ -0,0 +1,31 @@
...
+    assessment
+
+    trait :pending do
+      capital_assessment_result { 'pending' }
+    end
+
+    trait 'summarised' do
+      capital_assessment_result { 'summarised' }
+      lower_threshold { 3_000 }
+      upper_threshold { 8_000 }
```
reggieb: The thresholds are configurable - so likely to change and therefore should not be hard coded here. So this should be:
```ruby
 lower_threshold { Threshold.value_for(:capital_lower, at: Time.now) }
 upper_threshold { Threshold.value_for(:capital_upper, at: Time.now) }
```

spec/factories/capital_summary_factory.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    trait :at_lower_threshold do
+      assessed_capital { 3_000 }
+    end
+
+    trait :above_lower_threshold do
+      total_liquid { 3_000.01 }
+      total_capital { 3_000.01 }
+      assessed_capital { 3_000.01 }
+    end
```
reggieb: Again use the current thresholds and then:
```ruby
Faker::Number.between(from: 1, to: (threshold - 1)
```
and
```ruby
Faker::Number.between(from: (threshold + 1, to: (2 * threshold))
```
Or something similar so the numbers aren't hard coded, and random numbers are used to avoid the test being run on a number that just happens to be a special case.

db/schema.rb
stephenrichards: we can do without the not nulls I suppose, but if we don't have a default value, we have to set everything up or make sure we cope with nils.

spec/factories/capital_summary_factory.rb
stephenrichards: In a factory?  We don't want all the tests failing after a certain date just because the thresholds have changed.

app/controllers/capitals_controller.rb
```diff
@@ -36,7 +52,7 @@ def create
   def creation_service
     @creation_service ||= CapitalsCreationService.call(
       assessment_id: params[:assessment_id],
-      bank_accounts_attributes: input.dig(:liquid_capital, :bank_accounts),
+      bank_accounts_attributes: input[:liquid_capital],
```
Kalagan: `input[:liquid_capital]` is used here but the expected param is `bank_accounts` (line 23)
I'm surprised no tests are failing.

spec/factories/capital_summary_factory.rb
reggieb: If a change of threshold causes this to break that needs to occur in the tests. With hard coded thresholds that will not happen.

app/controllers/capitals_controller.rb
stephenrichards: Good catch.  Commenting out the testing of the format of the response also meant that we weretn't testing how many records were created.  Fizxed now.

app/services/base_workflow_service.rb
```diff
@@ -1,35 +1,12 @@
 class BaseWorkflowService
+  delegate :applicant, :capital_summary, to: :assessment
+  delegate :liquid_capital_items, :main_home, :additional_properties, :vehicles, to: :capital_summary
+
+  attr_reader :assessment
```
reggieb: All the attributes assigned in `initialize` should have a getter set, so the getters can be used in child objects rather than the instance variable. So:
```ruby
attr_reader :assessment, :submission_date, :calculation_period
```
The great advantage of this is that it makes it very obvious if you misspell one of these attributes in one of the child classes. For example, if you accidentally typed `@asssessment` this variable would just be assigned as nil and then passed to the next bit of code. If you used the getter and miss-typed it as `asssessment` you'd immediate get a `method not defined` error which is a lot easier to debug.

app/services/workflow_service/property_assessment.rb
```diff
@@ -1,84 +1,21 @@
 module WorkflowService
   class PropertyAssessment < BaseWorkflowService
     def call
+      @remaining_mortgage_allowance ||= Threshold.value_for(:property_maximum_mortgage_allowance, at: @submission_date)
```
reggieb: Defining `@remaining_mortgage_allowance` and then using it `calculate_property` unnecessarily couples the two methods. It would be simplest to remove this line and then update `calculate_property to:
```ruby
    def calculate_property
      remaining_mortgage_allowance = Threshold.value_for(:property_maximum_mortgage_allowance, at: @submission_date)
      Property.transaction do
        capital_summary.additional_properties.each do |property|
          property.assess_equity!(remaining_mortgage_allowance)
          remaining_mortgage_allowance -= property.allowable_outstanding_mortgage
        end
        capital_summary.main_home&.assess_equity!(remaining_mortgage_allowance)
      end
    end
```
And if you modify `BaseWorkflowService` as I've suggested, `@submission_date` should be replaced with `submission_date`

stephenrichards: Done

PR #119, opened 2019-12-10
AP-1103 Calculate gross income threshold

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1103)

Added gross income upper threshold data and a mechanism for deriving the correct threshold.

* new database column on `gross_income_summary` to hold `upper_threshold`.
* new values in `config/thresholds` to hold threshold-related data: 
 - `gross_income_upper` - the basic threshold
 - `dependant_increase_starts_after` - the number of dependants above which the threshold is increased
 - `dependent_step` - the amount each additional dependant adds to the threshold
 - `infinite_gross_income_upper` - where the threshold is waived, an 'infinite' threshold applies. This is encoded as £999,999,999,999
* new `GrossIncomeCollator` class to calculate the correct threshold an persist it to the database

NOTE: values for the 2020 gross income threshold are not yet known. The `config/threshold` file has been left blank for these. The approach used in `GrossIncomeCollator` to calculate the threshold works for the 2018/2019 thresholds. If the structure of the 2020 threshold is significantly different, some rework will be necessary.

## Checklist

### DIFF
--- a/app/services/workflow_service/gross_income_collator.rb
+++ b/app/services/workflow_service/gross_income_collator.rb
@@ -1,0 +1,39 @@
+module WorkflowService
+  class GrossIncomeCollator < BaseWorkflowService
+    def call
+      {
+        upper_threshold: upper_threshold
+      }
+    end
+
+    private
+
+    def upper_threshold
+      return infinite_threshold if assessment.matter_proceeding_type == 'domestic_abuse' && assessment.applicant.involvement_type == 'applicant'
+
+      Threshold.value_for(:gross_income_upper, at: assessment.submission_date) + dependant_increase
+    end
+
+    def infinite_threshold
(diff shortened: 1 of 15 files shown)


app/services/workflow_service/gross_income_collator.rb
```diff
@@ -0,0 +1,39 @@
...
+      @dependant_increase_starts_after ||= Threshold.value_for(:dependant_increase_starts_after, at: assessment.submission_date)
+    end
+
+    def dependant_step
+      @dependant_step ||= Threshold.value_for(:dependant_step, at: assessment.submission_date)
+    end
+
+    def number_of_dependants
+      assessment.dependants.count
+    end
```
stephenrichards: Not all dependants, just child dependants

PR #124, opened 2019-12-12
AP1100 State benefits endpoint

Obsiye:
State benefits endpoint

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1100)

Add StateBenefit model and migration file
Add StateBenefitPayment model and migration file

Add StateBenefitsCreationService, which will save the state benefits and payments into the state benefit table using methods on the StateBenefit model.

Update StateBenefitType column from :description to :name

Add spec test for state benefit model and request
Add factory for state benefit type
Update APIPIE documentation

Important:

Edit circleci config to not use db:setup for testing as it seeds the database. Instead to use db:create db:schema:load , which stops the database from being seeded as part of circleci tests.

To note:

If it's a unknown state benefit type passed in and is not in the StateBenefitType table, the StateBenefit model has code to make use of (reference) the 'Other' state benefit type in the state benefits table.

## Checklist

### DIFF
--- a/app/controllers/state_benefits_controller.rb
+++ b/app/controllers/state_benefits_controller.rb
@@ -1,0 +1,51 @@
+class StateBenefitsController < ApplicationController
+  resource_description do
+    short 'Add state benefits to an assessment'
+    formats ['json']
+    description <<-END_OF_TEXT
+    == Description
+      Adds details of an applicants'state benefits to an assessment.
+    END_OF_TEXT
+  end
+
+  api :POST, 'assessments/:assessment_id/state_benefit', 'Create state benefit'
+  formats ['json']
+  param :assessment_id, :uuid, required: true
+  param :state_benefits, Array, desc: 'Collection of state benefits' do
+    param :name, String, required: true, desc: 'The state benefit name'
+    param :payments, Array, desc: 'Collection of payment dates and amounts' do
+      param :date, Date, date_option: :today_or_older, required: true, desc: 'The date payment received'
+      param :amount, :currency, 'Amount of payment'
+    end
+  end
+
+  returns code: :ok, desc: 'Successful response' do
+    property :objects, array_of: Object
+    property :success, ['true'], desc: 'Success flag shows true'
+  end
+  returns code: :unprocessable_entity do
+    property :errors, array_of: String, desc: 'Description of why object invalid'
+    property :success, ['false'], desc: 'Success flag shows false'
+  end
+
+  def create
+    if creation_service.success?
+      render_success objects: creation_service.result
+    else
+      render_unprocessable(creation_service.errors)
+    end
+  end
+
+  private
+
+  def creation_service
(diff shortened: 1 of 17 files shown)


spec/models/state_benefit_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        expect(subject.state_benefit_type_id).to eq excluded_state_benefit_type.id
+      end
+      it 'name is blank' do
+        expect(subject.name).to be_blank
+      end
+    end
+    context 'other state benefit' do
+      before do
+        StateBenefitType.find_or_create_by(label: 'other', name: 'Other state benefit type', exclude_from_gross_income: false)
+      end
```
Obsiye: I used StateBenefitType.find_or_create_by in the StateBenefit model test
to add the state benefit type 'other' into state benefit types table, if it does not exist.
Cases when 'other' doesn't exist is when the db/seed isn't performed in testing, however, in circleci the seeds are run. There'll be a duplicate 'other' record which will cause an error as there is a uniqueness constraint on label in the state benefit types table.

app/controllers/state_benefits_controller.rb
```diff
@@ -0,0 +1,51 @@
+class StateBenefitsController < ApplicationController
+  resource_description do
+    short 'Add state benefits to an assessment'
+    formats ['json']
+    description <<-END_OF_TEXT
+    == Description
+      Adds details of an applicant'state benefit payments to an assessment.
```
stephenrichards: applicant's state benefits payments....

spec/requests/state_benefits_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    let!(:state_benefit_type_1) { create :state_benefit_type }
+    let!(:state_benefit_type_2) { create :state_benefit_type }
+
+    subject { post assessment_state_benefits_path(assessment_id), params: params.to_json, headers: headers }
+
+    context 'valid payload' do
+      context 'with two state benefits' do
+        it 'returns http success', :show_in_doc do
+          subject
```
stephenrichards: you need a `:show_in_doc` for an unsuccessful request as well so that we get to see what the error response is like in the documentation.

spec/models/state_benefit_spec.rb
stephenrichards: `find_or_create` in tests is a bit of a code smell - it means that you're never sure what state your DB is in when you run the tests.

I think we're better off writing  a factory to create this record and using it in a `let!` statement instead of the one we have in line 6

Obsiye: Thanks, I removed this. It was a hack around the test database being seeded in circleci. As we were using db:setup (which seeded the test database) so I changed it to 'db:create db:schema:load' in the circleci config file.

PR #126, opened 2019-12-13
AP-1136 - Implement passported assesment using workflow

stephenrichards:
## Use Workflow 
This PR inmplements workflow services to carry out the assessment, and changes the responsibility for carring out the sequence of events and updating the records from the CapitalSummary record to the workflow which is called by the AssessmentController

This will make it easier to handle themore complex workflows that will come with non-passported applicants

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1136)

## Checklist


spec/integration/test_runner_spec.rb
```diff
@@ -78,7 +78,7 @@ def verbosely_log(object, label)
     it 'process all worksheets and does not raise any error' do
       results = []
       worksheet_names.each do |spreadsheet_name|
-        # next unless spreadsheet_name == 'Test - P3 - Test1'
+        puts ">>>>>>>>> #{spreadsheet_name} #{__FILE__}:#{__LINE__} <<<<<<<<<<\n"
```
colinbruce: Debug statement left in...

colinbruce: Also, does spec/fixtures/integration_test_data.xlsx **need** to be an xlsx file?  
It needs specific software to edit/update it, could it be replaced with a CSV?

spec/support/attribute_matcher.rb
```diff
@@ -1,11 +1,16 @@
...
-    attribute_list.each do |meth|
-      return false unless actual_hash.__send__(meth) == expected_hash.__send__(meth)
-    end
+# returns true if all the attributes in expected hash have the same values in actual_model.
+# Does not care if there are extra attributees in the model which are not in the hash
+#
+RSpec::Matchers.define :have_matching_attributes do |expected_hash|
+  match do |actual_model|
+    actual_hash = actual_model.attributes.symbolize_keys
+    actual_hash.keep_if { |key, _vallue| expected_hash.key?(key) }
```
colinbruce: Typo in unused parameter, in both code blocks

spec/integration/test_runner_spec.rb
stephenrichards: These are now controlled by VERBOSE env var, set either to true for mildly verbose output, or noisy for very verbose output.  Nothing is printed if env var not set.

Yes, it does need to be ab .xlsx file because there is one tab for each test scenario, but it is created with Google sheets.

spec/support/attribute_matcher.rb
stephenrichards: 👍

PR #129, opened 2019-12-18
AP-1135 Non-passported workflow

stephenrichards:
## Non passported workflow

This PR implements the workflow for non-passported applicants.  Currently, it just calls the required collators and assessors in the right order, but some of the collators and assessors are yet to be written, so placeholder classes have been created for them

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1135)

- Removed `BenefitReceipt` model and table (replaced by `StateBenefitsPayment` model)
- Expanded `Workflows::NonPassportedWorkflow`
- Created `Workflows::SelfEmployedWorklflows` which for now will just raise an exception
- Added `self_employed` boolean to `Applicant` which is defaulted to false

## Checklist


spec/services/workflows/non_passported_workflow_spec.rb
```diff
@@ -2,12 +2,48 @@
...
+      context 'self_employed' do
+        let(:applicant) { create :applicant, self_employed: true }
+
+        it 'calls the self-employed workflow' do
+          expect(SelfEmployedWorkflow).to receive(:call).with(assessment)
+          subject
+        end
+      end
+
+      context 'not employed, not self_employed, Gross income exceeds threshold' do
```
Obsiye: Should not employed be a mock value from a factory method? As it's not explicit that the person is unemployed.

PR #134, opened 2019-12-20
AP-1120 Update State Benefit Type seeds with DWP codes

stephenrichards:
## Update Seed file.

We now have (a probably incomplete) list of DWP benefits and their codes to add to the list of benfits that are excluded from the means assessment by the LAA.  These new benefit types and their DWP codes have been added to the seeds

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1120)

- Added new field `dwp_code`

## Checklist

### DIFF
--- a/db/seeds/data/state_benefit_types.yml
+++ b/db/seeds/data/state_benefit_types.yml
@@ -196,6 +196,6 @@ winter_fuel_payment:
   :exclude_from_gross_income: false
   :dwp_code: WFP
 other:
-  :name: Other sate benefit
+  :name: Other state benefit
   :exclude_from_gross_income: false
   :dwp_code:
--- a/db/seeds/data/state_benefit_types.yml
+++ b/db/seeds/data/state_benefit_types.yml
@@ -3,75 +3,199 @@
 # name for example, from "Housing benefit" to "Homes benefit", change the 
 # name, but leave the benefit_type#label as it is.
 ---
-attendance_allowance:
-  name: Attendance Allowance
-  exclude_from_gross_income: true
-budget_advances:
-  name: Budget Advances
-  exclude_from_gross_income: true
-care_in_the_community_direct_payment:
-  name: Care in the Community Direct Payment
-  exclude_from_gross_income: true
-carers_allowance:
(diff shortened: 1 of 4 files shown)


db/seeds/data/state_benefit_types.yml
```diff
@@ -3,75 +3,199 @@
...
+  :exclude_from_gross_income: false
+  :dwp_code: WPA
+winter_fuel_payment:
+  :name: Winter Fuel Payment
+  :exclude_from_gross_income: false
+  :dwp_code: WFP
 other:
-  name: Other state benefit type
-  exclude_from_gross_income: false
+  :name: Other sate benefit
```
colinbruce: Typo on `State`

db/seeds/data/state_benefit_types.yml
```diff
@@ -3,75 +3,199 @@
...
+social_fund:
+  :name: Social Fund
+  :exclude_from_gross_income: false
+  :dwp_code: SF
 social_fund_payments:
-  name: Social Fund Payments
-  exclude_from_gross_income: true
+  :name: Social Fund Payments
+  :exclude_from_gross_income: true
+  :dwp_code:
```
colinbruce: Are Social Fund Payments known to be different from Social Fund? Should this be here without a code?

db/seeds/data/state_benefit_types.yml
stephenrichards: 👍 1

db/seeds/data/state_benefit_types.yml
stephenrichards: Yes it may be the same thing.  This is an amalgamation of two lists of benefits: one from the DWP with codes, the other from the LAA without codes.  The LAA may be slightly different wording for one of the DWP ones -it may be an entirely different benefit that isn't in the DWP list.

So we're going to use this table to identify ones with codes automatically from the bank statement.  We can also identify other DWP payments, and will ask the applicant about them and record the response.  We may be able in time to determine the codes for other benefits.

Dave is going to do a bit more research into the names the LAA use.

PR #177, opened 2020-02-11
AP-1212: Modify Housing Cost Calculator with cap

colinbruce:
## What

[Modify Housing Cost Calculator to take cap into account](https://dsdmoj.atlassian.net/browse/AP-1212)

## Checklist


spec/services/collators/housing_costs_collator_spec.rb
```diff
@@ -0,0 +1,329 @@
+require 'rails_helper'
+
```
stephenrichards: I think most of these tests should be in the `HousingCostCalculator` spec, the the spec for this class should just test that it calls the caclulator and updates the summary with the result.

PR #213, opened 2020-03-25
Expose State Benefit Type endpoint to return all types

Obsiye:
Expose State Benefit Type endpoint to return all types

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1322)

Add state benefit type GET route to return all state benefit types values for name, label and dwpcode.

Add a controller and request tests

IMPORTANT:

Updated the apipie for a GET request, which if the first for this application.

## Checklist

### DIFF
--- a/app/controllers/state_benefit_type_controller.rb
+++ b/app/controllers/state_benefit_type_controller.rb
@@ -4,17 +4,12 @@ class StateBenefitTypeController < ApplicationController
     formats ['json']
     description <<-END_OF_TEXT
     == Description
-      Returns all state benefit types.
+      Returns all state benefit types' name, label and dwpcode.
     END_OF_TEXT
   end
 
   api :GET, 'state_benefit_type', 'List of state benefit types'
   formats ['json']
-  param :state_benefit_type, Array, desc: 'Collection of state benefit types' do
-    param :name, String, required: false, desc: 'The state benefit type name'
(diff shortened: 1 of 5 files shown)


app/controllers/state_benefit_type_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  formats ['json']
+  param :state_benefit_type, Array, desc: 'Collection of state benefit types' do
+    param :name, String, required: false, desc: 'The state benefit type name'
+    param :label, String, required: false, desc: 'The state benefit type label'
+    param :dwp_code, String, required: false, desc: 'The state benefit type dwpcode'
+  end
+
+  def index
+    column_names = %w[name label dwp_code]
+    render json: StateBenefitType.pluck(column_names).map { |n, l, d| { name: n, label: l, dwp_code: d } }
```
stephenrichards: This is a bit cluncky, and I think the decision of how and what to render is probably best handled in the model.

So here I would just say:

<pre>render json: StateBenefitType.all.map(&:as_cfe_json)</pre>

and then on the model, have a method `#as_cfe_json` which just  calls the `#as_json` method on self, with an :only option specifying the fields you need.

PR #221, opened 2020-04-08
AP1348 Update dependant threshold allowances for 6th april

Obsiye:
[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1348)

AP1348 Update dependant threshold allowances for 6th April

The calculation_date/submission_date is already used for delegated and substantive application calculations.

Update tests.

To note:
The dependant type 'Partner' is not currently used so it is ignored.

## Checklist

### DIFF
--- a/spec/services/calculators/dependant_allowance_calculator_spec.rb
+++ b/spec/services/calculators/dependant_allowance_calculator_spec.rb
@@ -133,25 +133,25 @@ module Calculators
 
       describe 'child_under_15_allowance' do
         it 'returns the threshold value' do
-          expect(subject.child_under_15_allowance).to eq 291.49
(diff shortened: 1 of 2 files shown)


spec/services/calculators/dependant_allowance_calculator_spec.rb
```diff
@@ -133,25 +133,25 @@ module Calculators
```
stephenrichards, line 133: I think we need two sets of tests here - one with a submission date of 5th April 2020 which should return the old values, and one with a submission date after the 5th april that should return the new dates.

PR #222, opened 2020-04-14
AP-1361 Payment Period bugfix

stephenrichards:
## Fix bug that was causing the service to crash when given unorthodox payment periods

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1361)

There were in fact two bugs:
 - The `PaymentPeriodAnalyser` was not set up to handle arrays of dates comprising just one instance in the array
- Although the `MonthlyIncomeConverter` had been modified to just return the average monthly payment for unknown period types, the `MonthlyEquivalentCalculator` was raising an exception if the period type was unknown before calling it.

## Checklist

### DIFF
--- a/spec/services/utilities/payment_period_analyser_spec.rb
+++ b/spec/services/utilities/payment_period_analyser_spec.rb
@@ -210,6 +210,22 @@ module Utilities
           expect(described_class.new(dates).period_pattern).to eq :two_weekly
         end
       end
+
+      context 'irregular payments' do
+        context 'just one payment' do
+          let(:string_dates) { ['2019-05-15'] }
+          it 'returns two-weekly' do
+            expect(described_class.new(dates).period_pattern).to eq :unknown
+          end
+        end
+
+        context 'two payments 10 days apart' do
+          let(:string_dates) { '2019-05-15, 2019-05-25'.split(', ') }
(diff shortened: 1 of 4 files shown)


spec/services/utilities/payment_period_analyser_spec.rb
```diff
@@ -210,6 +210,22 @@ def verbose?
           expect(described_class.new(dates).period_pattern).to eq :two_weekly
         end
       end
+
+      context 'irregular payments' do
+        context 'just one payment' do
+          let(:string_dates) { ['2019-05-15'] }
+          it 'returns two-weekly' do
+            expect(described_class.new(dates).period_pattern).to eq :unknown
```
colinbruce, line 218: Should these two tests be ```it 'returns unknown```?

spec/services/utilities/payment_period_analyser_spec.rb
```diff
@@ -210,6 +210,22 @@ def verbose?
...
+          end
+        end
+
+        context 'two payments 10 days apart' do
+          let(:string_dates) { '2019-05-15, 2019-05-25'.split(', ') }
+          it 'returns two-weekly' do
```
colinbruce, line 224: see above

PR #224, opened 2020-04-21
AP-1374 monthly breakdown

stephenrichards:
## Implement Monthy income equivalents in response

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1374)

This PR implements the first part of AP-1374 - that is to include a structure monthly_income_equivalents in the response body.

- Limited types valid types of other income to `friends_or_family`, `maintenance_in`, `property_or_lodger`, `student_loan`, `pension` (or their humanized equivalents)
- Added validation to `OtherIncomeSource` records so that the nae is one of the above values
- Added extra columns to the `GrossIncomeSummary` record to hold the totals for each category
-  Added the extra section into the `GrossIncomeSummaryDecorator` to output the relevant data

## Checklist

### DIFF
--- a/app/models/other_income_source.rb
+++ b/app/models/other_income_source.rb
@@ -1,10 +1,13 @@
 class OtherIncomeSource < ApplicationRecord
   include MonthlyEquivalentCalculator
 
+  VALID_INCOME_SOURCES =  %w[friends_or_family maintenance_in property_or_lodger student_loan pension].freeze
+
   belongs_to :gross_income_summary
   has_many :other_income_payments
 
   validates :gross_income_summary_id, :name, presence: true
+  validates :name, inclusion: { in: VALID_INCOME_SOURCES }
 
   delegate :assessment, to: :gross_income_summary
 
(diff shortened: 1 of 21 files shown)


app/models/other_income_source.rb
```diff
@@ -1,10 +1,13 @@
 class OtherIncomeSource < ApplicationRecord
   include MonthlyEquivalentCalculator
 
+  VALID_INCOME_SOURCES = %w[friends_or_family maintenance_in property_or_lodger student_loan pension].freeze
+
   belongs_to :gross_income_summary
   has_many :other_income_payments
 
   validates :gross_income_summary_id, :name, presence: true
+  validates :name, inclusion: { in: VALID_INCOME_SOURCES }
```
Obsiye, line 10: A suggestion of adding a custom error message as the default will be "is not included in the list"
Is there a test for this, if needed?

`validates : name, inclusion: { in: VALID_INCOME_SOURCES, message: "%{value} is not a valid income source" }`

stephenrichards: Both good suggestions - I only saw this message after I merged it.  best to request changes.  I'll include it in the next ticket.

PR #233, opened 2020-04-30
Ap 1382 add deductions to the payload

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1382)

adds deductions (dependants_allowance and disregarded_state_benefits) to the disposable_income_summary payload

there is an existing dependants_allowance attribute, this will need to be removed once Apply has been amended to cope with moving it

## Checklist

### DIFF
--- a/app/models/state_benefit.rb
+++ b/app/models/state_benefit.rb
@@ -5,7 +5,7 @@ class StateBenefit < ApplicationRecord
   belongs_to :state_benefit_type
   has_many :state_benefit_payments
 
-  delegate :exclude_from_gross_income, to: :state_benefit_type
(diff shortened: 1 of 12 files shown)


app/models/state_benefit.rb
```diff
@@ -5,7 +5,7 @@ class StateBenefit < ApplicationRecord
   belongs_to :state_benefit_type
   has_many :state_benefit_payments
 
-  delegate :exclude_from_gross_income, to: :state_benefit_type
+  delegate :exclude_from_gross_income, :exclude_from_gross_income?, to: :state_benefit_type
```
Obsiye, line 8: little thing, you can ignore, I don't think this is needed ":exclude_from_gross_income" and we can keep ":exclude_from_gross_income?"

PR #236, opened 2020-05-06
AP-1399 Include monthly equivalents in integration tests

stephenrichards:
## Include monthly equivalents in integration tests

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1399)

The response structure from CFE has some extra sections in it 
- `monthly_income_equivalents`
- `monthly_outgoing_equivalents`
- `deductions`

This PR allows them to be compared and added into the result _**if they are present on the spreadsheet**_.  This allows the spreadsheets to be gradually updated while still allowing the specs to pass.

In noisy mode, fields which aren't tested (because they are blank either on the spreadsheet or in the response) are displayed in blue.

## Checklist


lib/integration_helpers/expected_result.rb
```diff
@@ -66,11 +91,11 @@ def verbose?
   end
 
   # :nocov:
-  def all_values_equal
-    methods.each do |method|
-      return false if __send__(method).to_s != @actual_result.__send__(method)
-    end
-  end
+  # def all_values_equal
```
willc-work, line 94: if this has been commented out can it be removed?

PR #270, opened 2020-06-16
AP1464 Add irregular income endpoint

Obsiye:
AP1464 Add irregular income endpoint

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1464)

- Create an IrregularIncomePayment model and table

- Add index to irregular_income_payments table to make sure income_type has one type e.g. student_loan per gross_income_summary

- GrossIncomeSummary has many IrregularIncomePayment

- Add a new field called :monthly_student_loan on the gross_income_summaries table.

- Add IrregularIncomeCreator to create irregular_income_payments

### Collators

- Update ChildcareCollator to use either other income source or irregular_income_payments for student loans

- Update GrossIncomeCollator to update :monthly_student_loan on the gross income summary. It's set to zero, if there is no irregular_income_payments

### Decorators

- Update **GrossIncomeSummaryDecorator** to include irregular_income into the json response

> Only include monthly_student_loan key in the GrossIncomeSummaryDecorator response when it's not in the other income sources. This prevents confusion with monthly_student_loan (from irregular_income_payments) and the student_loan in the other_income.
> Also, I left a TODO comment as this code will be updated to no longer use student_loan in other income and thus adding the key monthly_student_loan into the json response.

- Add **IrregularIncomePaymentsDecorator** to return [] if there are no irregular_income_payments or return the irregular income payments in the appropriate format

- Update **MonthlyIncomeEquivalentDecorator** to only include student_loan as a key in the response when there is no irregular_income_payments (which will handle student loan payments).

### Assessors

Update GrossIncomeAssessor test to include monthly_student_loan

### EXTRA:

- Update and create tests

- Update ApiPie documentation with Irregular Income

- Remove 'student_grant' and use 'student_loan' in other_income_source because 'student_grant' was not being used anywhere.

## Checklist

### DIFF
--- a/app/models/irregular_income_payment.rb
+++ b/app/models/irregular_income_payment.rb
@@ -1,0 +1,12 @@
+class IrregularIncomePayment < ApplicationRecord
+  ANNUAL_FREQUENCY = 'annual'.freeze
+  STUDENT_LOAN = 'student_loan'.freeze
+  VALID_IRREGULAR_INCOME_FREQUENCIES = [ANNUAL_FREQUENCY].freeze
+  VALID_IRREGULAR_INCOME_TYPES = [STUDENT_LOAN].freeze
+
+  belongs_to :gross_income_summary
+
+  validates :income_type, inclusion: { in: VALID_IRREGULAR_INCOME_TYPES }
+  validates :frequency, inclusion: { in: VALID_IRREGULAR_INCOME_FREQUENCIES }
+  validates :amount, numericality: { greater_than_or_equal_to: 0 }
+end
--- a/app/controllers/irregular_incomes_controller.rb
+++ b/app/controllers/irregular_incomes_controller.rb
@@ -1,0 +1,53 @@
+class IrregularIncomesController < ApplicationController
+  VALID_IRREGULAR_INCOME_TYPES = IrregularIncomePayment::VALID_IRREGULAR_INCOME_TYPES.freeze
+  VALID_IRREGULAR_INCOME_FREQUENCIES = IrregularIncomePayment::VALID_IRREGULAR_INCOME_FREQUENCIES.freeze
+
+  resource_description do
+    name 'Irregular income'
+    short 'Add irregular types of income details to an assessment'
+    formats ['json']
+    description <<-END_OF_TEXT
+    == Description
+      Adds details of an applicant's irregular income payments to an assessment.
+    END_OF_TEXT
+  end
+
+  api :POST, 'assessments/:assessment_id/irregular_income', 'Create irregular income'
+  formats ['json']
+  param :assessment_id, :uuid, required: true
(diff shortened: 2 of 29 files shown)


app/models/irregular_income_payment.rb
```diff
@@ -0,0 +1,23 @@
+class IrregularIncomePayment < ApplicationRecord
+  ANNUAL_FREQUENCY = 'annual'.freeze
+  STUDENT_LOAN = 'student_loan'.freeze
+  VALID_IRREGULAR_INCOME_FREQUENCY = [ANNUAL_FREQUENCY].freeze
```
stephenrichards, line 4: This is an array (even if it only has one item in it), so let's name it `VALID_IRREGULAR_INCOME_FREQUENCIES` in the plural.

app/models/irregular_income_payment.rb
```diff
@@ -0,0 +1,23 @@
...
+  VALID_IRREGULAR_INCOME_TYPES = [STUDENT_LOAN].freeze
+
+  belongs_to :irregular_income
+
+  validates_uniqueness_of :income_type
+  validates :amount, numericality: { greater_than_or_equal_to: 0 }
+
+  def student_loan?
+    return false unless annual?
+
```
stephenrichards, line 14: Do we need this here?  You're asking if it's a student loan, and responding false even if the income type is student loan but the frequency is not annual.  I'm not sure it's needed in any case - what is needed is validation on the income type and frequency.

app/models/irregular_income_payment.rb
```diff
@@ -0,0 +1,23 @@
+class IrregularIncomePayment < ApplicationRecord
+  ANNUAL_FREQUENCY = 'annual'.freeze
```
stephenrichards, line 2: I'm not sure we need this two-level approach here (of `IrregularIncome` and `IrregularIncomePayment`).  It is needed in some other types precisely because there are multiple payments of one type but here I'm anticipating that there will only be one student loan payment (and one of any other irregular payment types we may come up against in the future.  So I think we just need `GrossIncomeSummary` has many `IrregularIncomePayments` and no `IrregularIncome` in between.

app/models/irregular_income_payment.rb
```diff
@@ -0,0 +1,11 @@
...
+  STUDENT_LOAN = 'student_loan'.freeze
+  VALID_IRREGULAR_INCOME_FREQUENCIES = [ANNUAL_FREQUENCY].freeze
+  VALID_IRREGULAR_INCOME_TYPES = [STUDENT_LOAN].freeze
+
+  belongs_to :gross_income_summary
+
+  validates_uniqueness_of :income_type
+  validates :amount, numericality: { greater_than_or_equal_to: 0 }
+end
```
stephenrichards, line 11: You still have validates uniqueness of income type here, which means you will only ever be able to have one record in the database where income type is 'student_loan'.

let's just pair on this little bit straight after standup.

PR #281, opened 2020-07-03
AP-1441 Add client id to integration tests

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1441)

The integration test spreadsheet has been updated so that all transactions in the Other Incomes, Outgoings and State Benefits categories now include a `client_id`.

As a result, the `create_client_id` method in `DeeplyNestedPayloadGenerator` is no longer necessary. This commit removes it and adjusts the logic so that the `client_id` from the spreadsheet is used.

## Checklist

### DIFF
--- a/lib/integration_helpers/deeply_nested_payload_generator.rb
+++ b/lib/integration_helpers/deeply_nested_payload_generator.rb
@@ -28,8 +28,6 @@ class DeeplyNestedPayloadGenerator
     }
   }.freeze
 
-  FAKE_CLIENT_ID = SecureRandom.uuid.freeze
-
   def initialize(rows, payload_type)
     @rows = rows
     @payload_type = payload_type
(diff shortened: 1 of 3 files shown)


lib/integration_helpers/deeply_nested_payload_generator.rb
```diff
@@ -153,9 +153,4 @@ def store_client_id(row)
     _object, _source, _attr, value = row
     @payment_hash[client_id_method] = value
   end
-
-  # TODO: Remove #create_client_id method when integration tests spreadsheet passess in client_id with payments
-  def create_client_id(_row)
-    @payment_hash[client_id_method] = FAKE_CLIENT_ID
```
Obsiye, line 159: Looks good. Just a little leftover code. The point where this constant is defined "FAKE_CLIENT_ID" needs to be removed. On line 31 of this file.

mpw5: Thanks. Fixed!

PR #283, opened 2020-07-06
AP-1491 rewrite Frequency calculations

stephenrichards:
## Rewrite Frequency calculations

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1491)

The pre-existing frequency calculators were full of very complicated logic involving the mean, longest and shortest time between payment dates, the "slope" (i.e. whether the periods were getting longer or shorter) and then using various limitations to decide whether it was weekly, two weekly, four weekly or calendar-monthly.  We found that this was not always providing the correct response.

This has been completely rewritten using a different approach:
 - to see if the dates are exactly 7, 14, 28 days apart, or if they are the same calendar date each month
 - to see if any variation can be accounted for by paying 1 day early or 1 day late due to a bank holiday.

### Details
- Added a self-updating `BankHoliday` model to record all official bank holidays, getting data from GOV.UK
- Added the `business` gem to determine whether a particular date is a business day or not
- Replaced `Utilities::PaymentPeriodAnalyser` with `Utilities::NewPaymentPeriodAnalyser`
- Created class `Utilities::RegularPeriodAnalyser` to determine whether a series of dates if weekly, two-weekly or four-weekly after taking bank holidays into account
- Created class `Utilities::CalendarMonthlyPeriodAnalyser` to determine whether a series of dates is calendar monthly after taking holidays and weekends into account
- Removed old code:
  - `Calculators::TimeSeriesCalculator`
  - `Calculators::MonthlySalaryCalculator`
  - `Utilties::DateSlope`

##Checklist


app/services/utilities/new_payment_period_analyser.rb
```diff
@@ -0,0 +1,16 @@
+module Utilities
+  class NewPaymentPeriodAnalyser
```
colinbruce, line 2: Does this need to be prefixed *New*?
Just thinking bout future devs, I'd be wondering if there is an _original_ PaymentPeriodAnalyser

stephenrichards: yep - I'll change it.

PR #322, opened 2020-10-16
AP-1756 Add gem to implement content-security-policy headers

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1756)

See 4.14 of the security testing report.

Adds the `secure_headers` gem and configures it to generate content-security-policy (CSP) headers. The recommendation was to add this header `Content-Security-Policy: script-src 'self'; object-src 'none'; require-trusted-types-for 'script';` however as CFE is an API-only application I have set it at its most restrictive (ie `'none'` for all) as no scripts/styles/images should ever be loaded.

The suggested `require-trusted-types-for-` element is described in [guidance](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) as 'an experimental API that should not be used in production code' so I have omitted it.

## Checklist

### DIFF
--- a/config/initializers/secure_headers.rb
+++ b/config/initializers/secure_headers.rb
@@ -1,0 +1,8 @@
+SecureHeaders::Configuration.configure do |config|
+  # rubocop:disable Lint/PercentStringArray
+  config.csp = {
+    default_src: %w['none'],
+    script_src: %w['none']
+  }
+  # rubocop:enable Lint/PercentStringArray
+end
--- a/Gemfile
+++ b/Gemfile
@@ -48,6 +48,9 @@ gem 'google_drive'
 # parse spreadsheets
 gem 'roo', '~> 2.8.3'
 
+# Manage security headers
+gem 'secure_headers'
+
 group :development, :test do
   gem 'awesome_print'
   gem 'dotenv-rails', '>= 2.7.6'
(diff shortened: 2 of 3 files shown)


config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,8 @@
+SecureHeaders::Configuration.configure do |config|
+  # rubocop:disable Lint/PercentStringArray
+  config.csp = {
```
colinbruce, line 3: 🚧 not a blocker
As a matter of interest, why are we disabling rubocop rather than just changing the arrays to 
```ruby
%w[none]
```
?

colinbruce (approved):
Looks okay 👍 
One question that should not prevent merging

config/initializers/secure_headers.rb
mpw5: I wasn't too happy about that either. The gem needs the text to be in single quotes - if I run it without the quotes it throws: 

```default_src contains an invalid keyword source (none). This value must be single quoted. (SecureHeaders::ContentSecurityPolicyConfigError)```

The only way I could get it to play nicely with rubocop was by disabling the cop.

Comment (colinbruce) [reply config/initializers/secure_headers.rb:3]: I suppose the only other way would be to drop the array modifier completely and just use
```ruby
["'none'"]
```
But I'm not sure it's any better! 😃

PR #323, opened 2020-10-20
AP-1757 Update jQuery version that is a dependency in a gem

Obsiye:
### AP-1757 Update jQuery version that is a dependency in a gem

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1757)

- Update jQuery by adding the jquery downloaded file to the public folder. This overrides the jquery that is in the apipie gem which is out of date. This updates the jQuery version to 3.5.1 from 1.11.3 removing the security vunerability of an old dependency.

**To note:**

There is another ticket/task to use the yarn jQuery package and replace the above solution.

## Checklist


willc-work (approved):
LGTM

willc-work (approved):
👍

PR #328, opened 2020-10-26
BUGFIX Allow public file sources to be loaded for apipie

Obsiye:
Allow public file sources / resources to be loaded for apipie which wasn't because of the secure_headers csp change

After adding the secure_headers gem, assets from the public file were being blocked by the browser because it was not whitelisted. This has been seen with stylesheets and javascript not being loaded for apipie.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1780)

- Allow origin source to be whitelisted e.g. 'self' this is because we don't gather external resources

image

-----

image

-----
## Checklist


mpw5 (approved):
LGTM. Good catch 👍

PR #329, opened 2020-10-26
AP-1747: multi benefit flagging

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1747)

Add a new flagging system that allows Apply to send flagged values to CFE. These can then be parsed and, if required, flag the assessment that is passed back to Apply
This is designed to be extendable, i.e. in the future we can add more values into the `flags` field and check for their presence too
These can then be alerted to case workers when the assessment is returned and processed in Apply.

**NOTE**: this can be merged before the AP-1747 ticket on Apply but may need fixes. 
I am currently testing both ends of the ticket on a local machine but end-to-end testing on UAT may be tricky!

## Checklist

### DIFF
--- a/app/services/creators/state_benefits_creator.rb
+++ b/app/services/creators/state_benefits_creator.rb
@@ -43,10 +43,17 @@ module Creators
         state_benefit.state_benefit_payments.create!(
           payment_date: payment_params[:date],
           amount: payment_params[:amount],
-          client_id: payment_params[:client_id]
+          client_id: payment_params[:client_id],
+          flags: generate_flags(payment_params)
         )
       end
       state_benefit
+    end
+
+    def generate_flags(hash)
+      return false unless hash[:flags].present?
+
+      hash[:flags].map { |k, v| k if v.eql?(true) }.compact
     end
   end
 end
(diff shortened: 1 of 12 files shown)


app/services/creators/state_benefits_creator.rb
```diff
@@ -43,10 +43,17 @@ def create_state_benefit(state_benefit_params)
...
-          client_id: payment_params[:client_id]
+          client_id: payment_params[:client_id],
+          flags: generate_flags(payment_params)
         )
       end
       state_benefit
     end
+
+    def generate_flags(hash)
+      return false unless hash[:flags].present?
```
Obsiye, line 54: should it return nil?

Obsiye (approved):
LGTM! Just left a comment that wouldn't stop it from working

app/services/creators/state_benefits_creator.rb
colinbruce: I thought about nil  but, later in the process, we check for values and it seemed nicer for flags to be false.
To me, nil implied that we might have checked and it's false or something failed and we couldn't set a value.

Happy to discuss further if that doesn't make sense?

PR #339, opened 2020-11-06
update service.yaml

willc-work:
## What

Deploying with helm3 to master fails because we are trying to set clusterIP which sets a static IP address which we dont actually need to do. this should make it more flexible to prevent the failure

## Checklist


Obsiye (approved):
LGTM. Maybe test apply contacting its uat link before merging to master

stephenrichards (approved):

PR #342, opened 2020-11-09
Tech-debt: update ruby to 2.7

colinbruce:
## What

Update ruby to 2.7.2 to ensure we don't fall too far behind... v3 is only 2 months away!
- [x] update ruby to 2.7.1
- [x] fix warnings
- [x] document unfixable warnings
- [x] update ruby to 2.7.2

Ruby 2.7.2 hides the warnings that 2.7.0&1 highlighted.  There has been much discussion in the ruby community around it but that was my reason for making 2.7.1 as green as possible before moving to 2.7.2

## Checklist


spec/services/creators/other_incomes_creator_spec.rb
```diff
@@ -14,7 +14,14 @@ module Creators
       }
     end
 
-    subject { described_class.call(params) }
```
mpw5, line 17: Could `params` defined on line 8 be deleted if it's not being used here? (I can't see anywhere else where it's called but could be missing something)

mpw5 (approved):
LGTM 🎉  One question but not a blocker.

spec/services/creators/other_incomes_creator_spec.rb
colinbruce: Nice catch... I deleted the second one after replacing it and forgot to delete this one too!

PR #350, opened 2020-11-12
AP-1806 Fix flagging of childcare payments

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1806)

Childcare payments are being flagged for amount variation/frequency issues even if they have been disallowed from the calculation.

This change adds extra checks so that childcare is only flagged if the payments are included in the calculation, which only happens if the applicant is employed or in receipt of a student loan.

## Checklist

### DIFF
--- a/app/services/remark_generators/amount_variation_checker.rb
+++ b/app/services/remark_generators/amount_variation_checker.rb
@@ -1,13 +1,6 @@
 module RemarkGenerators
-  class AmountVariationChecker
-    def self.call(assessment, collection)
-      new(assessment, collection).call
-    end
-
-    def initialize(assessment, collection)
-      @assessment = assessment
-      @collection = collection
-    end
+  class AmountVariationChecker < BaseChecker
+    include Exemptable
 
     def call
       populate_remarks unless unique_amounts || exempt_from_checking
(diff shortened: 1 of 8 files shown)


app/services/remark_generators/amount_variation_checker.rb
```diff
@@ -28,5 +28,17 @@ def populate_remarks
...
+      childcare_payment? && childcare_disallowed?
+    end
+
+    def childcare_payment?
+      record_type == :outgoings_childcare
+    end
+
+    def childcare_disallowed?
+      @assessment.disposable_income_summary.childcare.zero?
+    end
```
colinbruce, line 42: Could these methods be extracted and shared between the two generators, either via a base_generator/concern/etc?

colinbruce (approved):
Looks good :+1:
One non-blocking observation for refactoring

app/services/remark_generators/amount_variation_checker.rb
mpw5: Excellent point. I had mental note to refactor but got overexcited and opened the PR too early 😁 

I've pushed a commit to change it if you fancy re-reviewing at some point.

colinbruce (approved):
💥  Nice 👍

PR #367, opened 2020-12-03
AP-1869 Policy change on property mortgage cap

Obsiye:
## AP-1869 Policy change on property mortgage cap

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1869)

1. Add configuration (config/thresholds/28-Jan-2021.yml) for when submission date is on or after 28th January 2021 - the policy change date. 

**To do the above:**

- Remove the mortgage cap by placing its value at 999_999_999_999, which subsequently draws back to using the outstanding mortgage instead of the fixed value of 100_000 in the property calculations.

- Add a new environment variable ALLOW_FUTURE_SUBMISSION_DATE to allow integration tests to run. Otherwise, they will fail because some of our tests in the means spreadsheet has submission dates in the future.

- Use the ALLOW_FUTURE_SUBMISSION_DATE in the date validator to decide whether or not to skip date validation on the submission date. There is :today_or_older validations elsewhere so created :submission_date_today_or_older

- Add tests

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,11 @@ PRIVATE_KEY_ID
 PRIVATE_KEY
 CLIENT_EMAIL
 CLIENT_ID
+ALLOW_FUTURE_SUBMISSION_DATE
 ```
+
+Set ALLOW_FUTURE_SUBMISSION_DATE to true to allow integration tests to run with submission dates that are in the future
+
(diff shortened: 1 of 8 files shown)


willc-work (approved):
Looks good👍

PR #377, opened 2020-12-16
AP-1914 Enable test threshold files for non-production environments

stephenrichards:
## Enable test threshold files for non-production environments

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1914)

This PR enables you to specify a test threshold file which will only be used in non-production environments

- Refactored `Threshold` so that it only loaded files specified in the `config/thresholds/values.yml` file, valid for the dates specified in that file.  This enables us to easily insert a new test file to be valid from any date.
- Enabled a new key `test_only` in the thresholds file, which if set to true, will prevent the file from loading unless the `USE_TEST_THRESHOLD_DATA` environment variable is set to `"true"`.
- updated the deploy files so that the USE_TEST_THRESHOLD_DATA env far was set to `"true"` for UAT and Staging, and `"false"` for production.

## Checklist


colinbruce (approved):
It's a fairly dense change, but I can't see anything obvious that worries me! 👍 

Would you consider refactoring and squashing the commits?  Future Colin would prefer a meaningful commit message rather than WIP 🙂

PR #386, opened 2021-01-04
BugFix Don't send exception object to Raven.capture message

Obsiye:
We capture api param errors with an array of string instead of an exception object for simplification, so that user
don't receive the error class in a response.

Solution is to have a seperate Raven.capture_exception when capturing param errors because render_unprocessable is used from child controllers to signify an internal error but without providing error class info etc.

This fixes the below (also found here [sentry](https://sentry.io/organizations/ministryofjustice/issues/2111912118/?project=5404900&query=is%3Aunresolved)):
```
NoMethodError

undefined method `backtrace' for #<Array:0x000055c94d4f9e00>
```

I believe this will stop CFE from sending a 500 error to a consumer (Apply app) and instead send the 422 :unprocessable_entity param error in json format whilst also sending a raven capture exception to alert the team.

## Checklist

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,11 +4,12 @@ class ApplicationController < ActionController::API
     render json: { success: false, errors: ["#{e.class}: #{e.message}"] }, status: :unprocessable_entity
   end
   rescue_from Apipie::ParamError do |e|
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
One comment around end point handling but, otherwise, LGTM :+1:

PR #398, opened 2021-01-18
AP-1921 cash transactions endpoint

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1921)

- Implement endpoint to accept details of cash income and outgoings and store in database
- moved a lot of the constants which were being repeated in different places into a module

## Checklist

### DIFF
--- a/app/controllers/cash_transactions_controller.rb
+++ b/app/controllers/cash_transactions_controller.rb
@@ -1,0 +1,65 @@
+class CashTransactionsController < ApplicationController
+
+  VALID_INCOME_CATEGORIES =  %w[benefits friends_or_family maintenance_in property_or_lodger pension].freeze
+  VALID_OUTGOING_CATEGORIES = %w[rent_or_mortgage child_care maintenance_out legal_aid].freeze
+
+  resource_description do
+    short 'Add cash transactions to an assessment'
+    formats ['json'] # rubocop:disable Layout/SpaceBeforeBrackets
+    description <<-END_OF_TEXT
+    == Description
+      Adds cash income and outgoings to and assessment so that they can be included in the means test.
+
+    END_OF_TEXT
+  end
+  api :POST, 'assessments/:assessment_id/cash_transactions', 'Add cash income and outgoings'
+  formats ['json'] # rubocop:disable Layout/SpaceBeforeBrackets
+  param :assessment_id, :uuid, required: true
+  param :income, Array, of: Hash, desc: 'Collection of Income categories' do
+    param :category, VALID_INCOME_CATEGORIES, required: true, desc: 'An identifying name for this income category'
+    param :payments, Array, of: Hash, desc: 'Total payments received for this category in each of the three months preceding the assessment date' do
+      param :date, Date, date_option: :today_or_older, required: true, desc: 'The date payment received (this must be the first day of one of the three months preceeding the assessment date'
+      param :amount, :currency, currency_option: :not_negative, required: true, desc: 'Amount of payment'
+      param :client_id, String, required: true, desc: 'Uniquely identifying string from client'
+    end
+  end
+  param :outgoings, Array, of: Hash, desc: 'Collection of Outgoing categories' do
+    param :category, VALID_OUTGOING_CATEGORIES, required: true, desc: 'An identifying name for this outgoing category'
+    param :payments, Array, of: Hash, desc: 'Total payments made for this category in each of the three months preceding the assessment date' do
+      param :date, Date, date_option: :today_or_older, required: true, desc: 'The date payment made (this must be the first day of one of the three months preceeding the assessment date'
+      param :amount, :currency, currency_option: :not_negative, required: true, desc: 'Amount of payment'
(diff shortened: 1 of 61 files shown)


app/services/bank_holiday_retriever.rb
```diff
@@ -28,7 +26,7 @@ def response
   end
 
   def uri
-    URI.parse(API_URL)
+    URI.parse(CFEConstants::GOVUK_BANK_HOLIDAY_API_URL)
```
Obsiye, line 29: I think we should keep what we already have here or maybe it should go in an environment variable.

There's a dependancy between the two constants
API_URL and DEFAULT_GROUP but CFEConstants won't be able to show that clearly.

I wonder if this should be renamed as GovUkBankHolidayRetriever

app/controllers/cash_transactions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+      param :amount, :currency, currency_option: :not_negative, required: true, desc: 'Amount of payment'
+      param :client_id, String, required: true, desc: 'Uniquely identifying string from client'
+    end
+  end
+  param :outgoings, Array, of: Hash, desc: 'Collection of Outgoing categories' do
+    param :category, CFEConstants::VALID_OUTGOING_CATEGORIES, required: true, desc: 'An identifying name for this outgoing category'
+    param :payments, Array, of: Hash, desc: 'Total payments made for this category in each of the three months preceding the assessment date' do
+      param :date, Date, date_option: :today_or_older, required: true,
+                         desc: 'The date payment made (this must be the first day of one of the three months preceeding the assessment date'
+      param :amount, :currency, currency_option: :not_negative, required: true, desc: 'Amount of payment'
```
Obsiye, line 28: if it's an outgoing payment, shouldn't it allow negative payments? Instead it shouldn't be more than zero?
or at least not  have that currency_option validation

It depends on how we want the data if we want outgoings: { payment: -100.00 } or  outgoings: { payment: 100 }

spec/models/cash_transaction_category_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      it 'is invalid' do
+        expect(category).not_to be_valid
+        expect(category.errors[:name]).to eq ["can't be blank"]
+      end
+    end
+
+    context 'operation credit' do
+      let(:operation) { 'credit' }
+      context 'invalid name' do
+        let(:name) { 'housing_costs' }
```
Obsiye, line 22: should these :name attributes in this test come from CFEConstants::VALID_INCOME_CATEGORIES or  CFEConstants::VALID_OUTGOING_CATEGORIES rather than be manually input?

app/services/creators/cash_transactions_creator.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    def validate_category(category_hash)
+      if category_hash[:payments].size != 3
+        @errors << "There must be exactly 3 payments for category #{category_hash[:category]}"
+        return
+      end
+      validate_payment_dates(category_hash)
+    end
+
+    def validate_payment_dates(category_hash)
```
Obsiye, line 64: could #validate_payment_dates be tested, please, especially around `unless dates == valid_dates `

stephenrichards (commented):
For some reason, I can't comment on Abd's comments, but regarding the comment on `spec/models/cash_transaction_category_spec.rb`, no I dont think we should use the constants here, because in a way, that's what we're testing.

app/services/creators/cash_transactions_creator.rb
stephenrichards: It is tested here: https://github.com/ministryofjustice/check-financial-eligibility/blob/c4283889e259bbe1117b162eb2f4d5bae35d6ba4/spec/services/creators/cash_transactions_creator_spec.rb#L75

Obsiye (approved):
LGTM!

app/services/creators/cash_transactions_creator.rb
```diff
@@ -63,7 +60,19 @@ def validate_category(category_hash)
...
-      @errors << "Expecting payment dates for category #{category_hash[:category]} to be #{pretty_dates}" unless dates == valid_dates
+      return if dates == first_three_valid_dates
+
+      return if dates == last_three_valid_dates
+
+      @errors << "Expecting payment dates for category #{category_hash[:category]} to be 1st of three of the previous 3 months"
+    end
+
+    def first_three_valid_dates
+      valid_dates.slice(0, 3)
```
Obsiye, line 71: should it be `valid_dates.slice(0, 2)`
so it's the first three

Alex-Swann: @Obsiye hiyah, no, the first argument stipulates the index to start from, whereas the second argument indicates the number of objects to slice from the array rather than the index number to slice to. It can look like `(start_index, end_index)` whereas it is `(start_index, total)`
e.g.
```
irb(main):040:0> arr = [0,1,2,3]
irb(main):041:0> arr.slice(2,1)
=> [2]
```

PR #410, opened 2021-01-29
AP-1983: Add bank, cash and all_sources for disposable incomes for assessment version 3

Alex-Swann:
## What

[AP-1983](https://dsdmoj.atlassian.net/browse/AP-1983)

Add new fields to disposable income summary table and ensure formatting correctly in json response for assessment version 3 only.

## Checklist

### DIFF
--- a/spec/services/calculators/housing_costs_calculator_spec.rb
+++ b/spec/services/calculators/housing_costs_calculator_spec.rb
@@ -282,6 +282,293 @@ module Calculators
       end
     end
 
+    context 'version 3 including cash payments' do
+      let(:assessment) { create :assessment, :with_v3, :with_disposable_income_summary, with_child_dependants: children }
+      let(:rent_or_mortgage_category) { assessment.cash_transaction_categories.detect { |cat| cat.name == 'rent_or_mortgage' } }
+      let(:rent_or_mortgage_transactions) { rent_or_mortgage_category.cash_transactions.order(:date) }
+      let(:monthly_cash_housing) { rent_or_mortgage_transactions.average(:amount).round(2).to_d }
+
+      before do
+        subject
+        assessment.disposable_income_summary.reload
+        @assessment = assessment
+      end
+
+      context 'when applicant has no dependants' do
+        let(:housing_cost_amount) { 1200.00 }
+
+        context 'and does not receive housing benefit' do
+          context 'board and lodging' do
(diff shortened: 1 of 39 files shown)


spec/services/calculators/housing_costs_calculator_spec.rb
```diff
@@ -282,6 +284,294 @@ module Calculators
...
+              end
+            end
+          end
+        end
+      end
+    end
+
+    def monthly_cash_housing
+      @monthly_cash_housing ||= monthly_transaction_amount_by(operation: :debit, category: :rent_or_mortgage)
+    end
```
stephenrichards, line 573: is this a valid test if it is using the live codes `monthly_transaction_amount_by` method to come up with a figure that it plugs into the test, and then test if it is that figure?

spec/services/collators/gross_income_collator_spec.rb
```diff
@@ -130,21 +130,9 @@ module Collators
...
 
         before do
-          3.times do
-            create :cash_transaction, cash_transaction_category: benefits_in_cash
-            create :cash_transaction, cash_transaction_category: friends_or_family_in_cash
-            create :cash_transaction, cash_transaction_category: maintenance_in_cash
-            create :cash_transaction, cash_transaction_category: property_or_lodger_in_cash
-            create :cash_transaction, cash_transaction_category: pension_in_cash
-          end
           subject
```
stephenrichards, line 136: Nice moving this into the factory, where it should have been in the first place. mea culpa.

stephenrichards (changes requested):

spec/services/calculators/housing_costs_calculator_spec.rb
Alex-Swann: mmm, not sure. The test amounts created are a random figure between 10-100. What they are we don't care about. As they are randomly generated I thought this the easiest way to grab them. The tests above add the monthly cash amount to the monthly bank amounts, if they didn't exist or were `0` then the relevant caps wouldn't be hit above and the tests would fail. I've changed the hard coded bank amounts to accommodate for this. The bank amounts are in the outgoings table and easier to access than the cash transactions as they link up through two tables, so to avoid additional application code for just the sake of the test suite I thought it simpler to do it this way.

stephenrichards (approved):

PR #450, opened 2021-03-18
AP-2111 format error messages instead of crashing

stephenrichards:
## format error messages instead of crashing

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2111)

CFE is crashing instead of formatting a proper error message and sending it to Sentry.  This is the first fix that will format an array of error messages as a string so that Sentry can display it properly.  Then we will be able to see what the cause of the error is and fix that.

## Checklist


willc-work (approved):
Looks good

PR #471, opened 2021-04-09
AP-2175 Threshold waiver by proceeding type

stephenrichards:
Once we switch to V4 of the API, we will need to know whether certain thresholds are waived for a particular proceeding type.

This PR provides the `ProceedingTypeThreshold` class to do that.

If it is one of the thresholds which are subject to a waiver, and if the proceeding type is one where that threshold is waived, then the infinite upper limit of 999,999,999,999 is returned, otherwise Threshold is called to get the normal value.

Call like:
 ```
    ProceedingTypeThreshold.value_for(:DA004, :capital_upper, calculation_date)
```

## What

## Checklist

### DIFF
--- a/app/models/proceeding_type_threshold.rb
+++ b/app/models/proceeding_type_threshold.rb
@@ -1,0 +1,104 @@
+# This class returns the correct threshold for a specified proceeding type
+#
+class ProceedingTypeThreshold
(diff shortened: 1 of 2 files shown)


Obsiye (approved):
LGTM

PR #474, opened 2021-04-13
TESTFIX - fix dates on tests 

stephenrichards:
## Fix tests on dates.

Some unit tests have been written using today's dates, and expecting the values to be caluclated according to the current values in the config files.  When the thresholds and values change, we have to change the expected values in the dates.

This fixes the date for the test so that we don't have to do that every time there is a change to he means test.

## Checklist


Obsiye (approved):
LGTM

PR #493, opened 2021-04-29
AP-2251 output response in V4 format for V4 assessments

stephenrichards:
## Output response in V4 format for V4 assessments

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2251)

- Move all existing decorators to the Decorators::V3 namespace
- Create all new V4 decorators in Decorators::V4 namespace
- Decide which decorator class to use based on version attribute on assessment
- Add `Eligibility::Assessment` class to record overall eligibility for all proceeding types

## Checklist


app/controllers/assessments_controller.rb
```diff
@@ -63,15 +63,16 @@ def show
     perform_assessment
   rescue StandardError => err
     Sentry.capture_exception(err)
-    render json: Decorators::ErrorDecorator.new(err).as_json, status: :unprocessable_entity
+    render json: Decorators::V3::ErrorDecorator.new(err).as_json, status: :unprocessable_entity
```
colinbruce, line 66: Will check as review continues... do we need a 
```ruby
Decorators::V4::ErrorDecorator
```
too?

app/models/assessment.rb
```diff
@@ -37,21 +42,29 @@ def remarks
...
+    return if version_3?
 
     proceeding_type_codes.each do |code|
       errors.add(:proceeding_type_codes, "invalid: #{code}") unless code.to_sym.in?(ProceedingTypeThreshold.valid_ccms_codes)
     end
   end
 
   def proceeding_types_codes_required?
-    version != '3'
+    version_4?
```
colinbruce, line 68: Is this too fragile? Will it break on the inevitable v5?  Should we have a `version>3?` method?

spec/services/assessors/main_assessor_spec.rb
```diff
@@ -2,117 +2,15 @@
...
-        expect(%w[pending ineligible ineligible]).to have_assessment_error(assessment, 'Assessment not complete: Gross Income assessment still pending')
-        expect(%w[pending ineligible pending]).to have_assessment_error(assessment, 'Assessment not complete: Gross Income assessment still pending')
-        expect(%w[pending pending contribution_required]).to have_assessment_error(assessment, 'Assessment not complete: Gross Income assessment still pending')
-        expect(%w[pending pending eligible]).to have_assessment_error(assessment, 'Assessment not complete: Gross Income assessment still pending')
-        expect(%w[pending pending ineligible]).to have_assessment_error(assessment, 'Assessment not complete: Gross Income assessment still pending')
-        expect(%w[pending pending pending]).to have_assessment_error(assessment, 'Assessment not complete: Gross Income assessment still pending')
-      end
-    end
+    # TODO: write tests here that look at all the assessment_eligibility results and resturn eligible, ineligible, partly eligible
```
colinbruce, line 14: Should these have been written or is it a longer term aim?

spec/services/decorators/v4/disposable_income_result_decorator_spec.rb
```diff
@@ -0,0 +1,90 @@
+require 'rails_helper'
+
+module Decorators
+  module V3
```
colinbruce, line 4: File and module name mismatch, V4 but module still v3

colinbruce (changes requested):

app/controllers/assessments_controller.rb
stephenrichards: I don't think we do - the error format is identical, so we can just use that one. I suppose it might make sense to move it out of the V3 namespace and just call it `Decorators::ErrorDecorator` again.

spec/services/assessors/main_assessor_spec.rb
stephenrichards: Oh no, I still need to write thse tests I think.

spec/services/decorators/v4/disposable_income_result_decorator_spec.rb
stephenrichards: 👍

colinbruce (approved):

PR #505, opened 2021-05-10
AP-2252 multiple worksheets & V4 comparison

stephenrichards:
## Multiple worksheets & V4 comparison

This PR introduces two major changes to the way in which integration specs are run:

1. Enable Multiple spreadsheets to be processed
In order to enable similar tests to be grouped together, the integration test will now process multiple spreadsheets. The master spreadsheet named `AAA - CFE Integration Test master spreadsheet` has three tabs, the first of which lists the spreadsheets that are to be processed. (the other two tabs contain the skeleton spreadsheets for V3 and V4 tests).

2. Enable both V3 and V4 format spreadsheets to be processed
Each spreadsheet has now been altered to contain the version number.  If version 3, then the existing format is expected, if version 4, then matter type is not specified, but a semi-colon list of proceeding type codes is specified.  The results format is different too, giving results per matter type, and per proceeding type.

### Overview of Code changes

- Environment variable values placed into Rails configuration stucture to comply with new rubocop rules
- New Class `TestCase::GroupRunner` to read the master spreadsheet, load the test cases and pass each one to `TestCase::Worksheet` for processing
- Existing class `TestCase::ExpectedResult` moved to `TestCase::V3::ExpectedResult`
- New class `TestCase::V4::ExpectedResult` to parse the result section of V4 spreadsheets
- New class 'TestCase::V4::ResultComparer` to compare the actual and expected results of V4 tests 

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2252)

## Checklist


willc-work (approved):
Looks good

PR #542, opened 2021-07-05
AP-2381 Migrate Integration test spreadsheets to use V4

stephenrichards:
## Migrate Integration test spreadsheets to use V4

This PR adds a class `IntegrationTestMigrator` which reads all existing V3 spreadsheets and creates a V4 style CSV.

I have imported those CSVs into the appropriate spreadsheet, and we can remove all the V3 tests at a later date.

This PR does not affect the way in which CFE works in any way.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2381)

## Checklist


Obsiye (approved):
I've read as much as I can into this. However, I can't assure it won't break. 

I think this would work best when having the new spreadsheet up then running V4 integration tests against it.

PR #602, opened 2021-09-15
Tech-Debt: Flickering decimal tests

colinbruce:
## What

There have been some more flickers of decimal creation and testing and this PR fixes the latest one and attempts to identify and correct any others that may occur in the tests

This issue in `housing_costs_calculator_spec.rb` seems to be caused by changing one side of the equation, the occasional decimal flicker would cause issues with the float comparison.  After running multiple tests, ensuring all values are either comparing decimals with decimals or floats with floats seems to resolve it 🤞 

## Checklist


mpw5 (approved):
LGTM!

PR #663, opened 2021-11-25
AP-2700 No longer deploy to old UAT instance

stephenrichards:
## No longer deploy to old UAT instance

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2700)

## Checklist


willc-work (commented):
Could you rename the file: deploy/helm/values-live-uat.yaml to just be values-uat.yaml

This would also mean changing the following line in bin/uat_deploy_to_live

--values ./deploy/helm/values-live-uat.yaml \

And the file uat_deploy_to_live could also be renamed to just uat_deploy

Not blockers but this would keep it consistent across the services.

mpw5 (approved):
LGTM

PR #664, opened 2021-11-25
Call employment income calculator on non-passported workflow

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2671)

Call employment_income_calculator on non-passported workflow.

## Checklist

### DIFF
--- a/app/services/collators/gross_income_collator.rb
+++ b/app/services/collators/gross_income_collator.rb
@@ -5,6 +5,7 @@ module Collators
     INCOME_CATEGORIES = CFEConstants::VALID_INCOME_CATEGORIES.map(&:to_sym)
 
     def call
+      Calculators::EmploymentIncomeCalculator.call(assessment) if assessment.employments.any?
       gross_income_summary.update!(populate_attrs)
     end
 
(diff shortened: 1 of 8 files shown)


app/services/collators/gross_income_collator.rb
```diff
@@ -14,15 +15,21 @@ def populate_attrs
...
 
       attrs
     end
 
+    def populate_income_attrs(attrs, category)
+      attrs[:"#{category}_bank"] = category == :benefits ? monthly_state_benefits : categorised_income[category].to_d
+      attrs[:"#{category}_cash"] = monthly_transaction_amount_by(operation: :credit, category: category)
+      attrs[:"#{category}_all_sources"] = attrs[:"#{category}_bank"] + attrs[:"#{category}_cash"]
+      attrs[:total_gross_income] += attrs[:"#{category}_all_sources"]
+      attrs
```
mpw5, line 30: I'm not sure you need to return `attrs` here, it doesn't look like the return value from this method is being used anywhere. (I could be wrong though, I haven't tried it out).

mpw5 (approved):
LGTM. I've left one comment, but it's not a blocker.

app/services/collators/gross_income_collator.rb
kmahern: thanks @mpw5 , changed.

PR #685, opened 2021-12-21
AP-2683: Remove openstruct

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2683)

Remove the only remaining use of OpenStruct in the service by:
* Remove the open_structify helper method, it is the only code to call DatedStruct and is no-longer referenced in the codebase
* Remove the DatedStruct class , this honours the `remove this method (or even the whole class) once all reference to it has been removed from other old_services` TODO as it was only referenced in the now deleted open_structify helper_method
* Remove the rubocop Style/OpenStructUse exemption for DatedStruct now we have removed the DatedStruct class

## Checklist


mpw5 (approved):
LGTM

PR #744, opened 2022-01-20
AP-2770 Update ingress.yaml to use networking.k8s.io/v1

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2770)

Updates `ingress.yaml` to use `networking.k8s.io/v1` as a result of moving to the live cluster. This reconfigures the ingress using the new syntax.

## Checklist


colinbruce (approved):
LGTM :+1:

PR #770, opened 2022-01-31
More GOVUK Rubocop todos

stephenrichards:
## More GOVUK Rubocop todos

## Checklist


spec/services/utilities/calendar_monthly_period_analyser_spec.rb
```diff
@@ -191,11 +193,12 @@ module Utilities
             end
           end
 
-          context "payment date is after holiday"
-          let(:dates) { make_dates(%w[2019-05-09 2019-06-09 2019-07-10]) }
+          context "payment date is after holiday" do
+            let(:dates) { make_dates(%w[2019-05-09 2019-06-09 2019-07-10]) }
```
mpw5, line 197: It looks like the same value for `dates` is now being created on lines 186 and 197. I think one of them can probably be removed.

mpw5 (approved):
Looks good. One non-blocking comment.

spec/services/utilities/calendar_monthly_period_analyser_spec.rb
stephenrichards: Good spot.  Fixed.

PR #782, opened 2022-02-04
AP-2850: Update Ingress hosts

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2850)

After the EKS migration, we ended up with a variety of legacy and modern hosts on all of our services.
We agreed on a new pattern to be applied to all services in the apply namespace
A future commit will remove the legacy hosts

## Checklist


mpw5 (approved):
Looks good 👍 

I noticed you've left the `.apps.live.cloud-platform.service.justice.gov.uk` version of the url in the staging and production values files. Is there any reason for that? I haven't included them in the HMRC change 😬

colinbruce:
> 

Just in case someone had updated their local dev environment 🤷‍♂️ 
Belt and braces, but probably not necessary - I think your approach is fine too

PR #807, opened 2022-02-18
AP-2892 Employment Income calculation

stephenrichards:
## Employment Income calculation

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2892)

Implemented full employment calculation, taking into account variation in payments, and tax and NI refunds.

- when more than one employment record, set values to zero and add remark
- When monthly equivalent income of payments varies by less than £60, the most recent payment is taken as the monthly equivalent to use.  Above that, the blunt average is used.
- Tax and NI refunds are treated as zero
- Tax and NI monthly equivalents are calculated using whatever method was used for gross income

Changes:
- Added new configuation variable `employment_income_variance` to define the limit of variance between the two calculation methods.
- Added `gross_income_monthly_equiv` column to the `employment_payments` table - this holds the gross_income figure converted to a monthly equivalent, and is required because the treatment of varying amounts depends on the difference between the highest and lowest monthly equivalent payments
- Added `calculation_method` column to `employments` to record which of the two calculation methods was used, so that the same can be used for NI and Tax
- Created 'calculate!' method on `Employment` model to carry out the required calculations
- Created class `Calculators::MultipleEmploymentsCalculator` to take care of zeroing values when multiple employments
- Created class `Calculators::TaxNiRefundCalculator` to zero out any tax or NI refunds
- Altered `GrossIncomeDecorator` to always output JSON for employments in order of Job name (essential for testing)
- Created class `Utilties::EmploymentIncomeMonthlyEquivalentCalculator` to calculate monthly equivalent from payment amount and frequency
- Created class `Utilties::EmploymentIncomeVariationChecker` to check to see if variation is above or below threshold

## Checklist


app/services/remarks.rb
```diff
@@ -5,20 +5,6 @@ class Remarks
   # TODO:
   # Remove the above types when no longer required
```
colinbruce, line 6: Can the comment go too?

colinbruce (approved):
Only a minor observation around comments, but otherwise LGTM :+1:

PR #841, opened 2022-03-17
BUGFIX: display expected values for integration tests

willc-work:
## What
the integration tests were not displaying the expected values for NI/TAX/fixed_emp_allowance, as it was looking for these in `gross_income_summary`, however they exist in `disposable_income_summary`.

This change displays the value and if it is a positive value it makes it a negative in line with the actual values CFE outputs


lib/integration_helpers/test_case/v4/result_comparer.rb
```diff
@@ -323,15 +323,15 @@ def expected_employment_income_benefits_in_kind
       end
 
       def expected_employment_income_tax
-        expected_gross_income[:employment_income_tax]
+        expected_disposable_income[:employment_income_tax]&.positive? ? -expected_disposable_income[:employment_income_tax] : expected_disposable_income[:employment_income_tax]
```
stephenrichards, line 326: I don't think this is right - the code ALWAYS makes it a negative figure.  But if they are getting a tax (or NI) refund, then it will be a positive figure and show as such I t hink.

stephenrichards (changes requested):

PR #860, opened 2022-03-30
Ap 2982 show actual disposable income

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2982)

- Allow negative total disposable incomes
- Amend pensioner_capital_disregard monthly_income_values in threshold files to cater for negative values
- Amend disposable_income_contribution_bands band_zero on threshold files to allow negative values

## Checklist

### DIFF
--- a/config/thresholds/2018-04-08.yml
+++ b/config/thresholds/2018-04-08.yml
@@ -13,7 +13,7 @@ pensioner_capital_disregard:
   non_passported: 100_000
   # capital to disregard if monthly disposable income (excluding income derived from capital) is below these figures
   monthly_income_values:
-    [-999_999_999_999, 25.99]: 100_000
+    [-.inf, 25.99]: 100_000
     [26.0, 50.99]: 90_000
     [51.0, 75.99]: 80_000
     [76.0, 100.99]: 70_000
@@ -43,7 +43,7 @@ single_monthly_housing_costs_cap: 545
 
 disposable_income_contribution_bands:
   band_zero:
-    threshold: 0.0
+    threshold: -.inf
     disregard: 0.0
     percentage: 0.0
     base: 0.0
--- a/config/thresholds/2018-04-08.yml
+++ b/config/thresholds/2018-04-08.yml
@@ -13,7 +13,7 @@ pensioner_capital_disregard:
   non_passported: 100_000
   # capital to disregard if monthly disposable income (excluding income derived from capital) is below these figures
   monthly_income_values:
-    [0, 25.99]: 100_000
+    [-999_999_999_999, 25.99]: 100_000
     [26.0, 50.99]: 90_000
     [51.0, 75.99]: 80_000
     [76.0, 100.99]: 70_000
--- a/spec/services/calculators/income_contribution_calculator_spec.rb
+++ b/spec/services/calculators/income_contribution_calculator_spec.rb
@@ -13,6 +13,14 @@ module Calculators
         end
       end
 
+      context "negative income" do
+        let(:income) { -100.0 }
+
+        it "returns zero" do
+          expect(calculator).to be_zero
+        end
+      end
+
       context "income in band a" do
         let(:income) { 340.0 }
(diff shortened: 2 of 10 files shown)


config/thresholds/2018-04-08.yml
```diff
@@ -13,7 +13,7 @@ pensioner_capital_disregard:
   non_passported: 100_000
   # capital to disregard if monthly disposable income (excluding income derived from capital) is below these figures
   monthly_income_values:
-    [0, 25.99]: 100_000
+    [-.inf, 25.99]: 100_000
```
jsugarman, line 16: 👍  interesting

spec/services/calculators/income_contribution_calculator_spec.rb
```diff
@@ -13,6 +13,14 @@ module Calculators
         end
       end
 
+      context "negative income" do
+        let(:income) { -100.0 }
+
+        it "returns zero" do
+          expect(calculator).to be_zero
+        end
```
jsugarman, line 21: nit
```suggestion
        it { expect(calculator).to be_zero }
```

spec/services/collators/disposable_income_collator_spec.rb
```diff
@@ -69,6 +69,19 @@ module Collators
...
+        before do
+          assessment.gross_income_summary.update!(total_gross_income: total_outgoings - 1500.0)
+        end
+
+        it "returns the correct negative amount" do
+          collator
+          result = assessment.gross_income_summary.total_gross_income + assessment.gross_income_summary.gross_employment_income - disposable_income_summary.reload.total_outgoings_and_allowances
+          expect(disposable_income_summary.total_disposable_income).to eq result
+          expect(disposable_income_summary.total_disposable_income).to be_negative
+        end
```
jsugarman, line 82: nit is it possible to break this up.

```suggestion

          let(:result) do 
            assessment.gross_income_summary.total_gross_income + 
            assessment.gross_income_summary.gross_employment_income - 
            disposable_income_summary.reload.total_outgoings_and_allowances
          end
          
          before { collator }
          
        it "returns the correct amount" do
          expect(disposable_income_summary.total_disposable_income).to eq result
        end
        
        it "can be a negative amount" do
          expect(disposable_income_summary.total_disposable_income).to be_negative
        end
```

jsugarman: I have not tried this though

jsugarman: @kmahern have created PR https://github.com/ministryofjustice/check-financial-eligibility/pull/861 which breaks this up if you like. but not a blocker

jsugarman (approved):
👍  approving based on my [limited] understanding of the codebase but have submitted this separate pr, https://github.com/ministryofjustice/check-financial-eligibility/pull/861, to handle one of my comments if you like.

spec/services/collators/disposable_income_collator_spec.rb
colinbruce: Since we started following the GOVUK Rubocop style guide, they [have a note](https://github.com/alphagov/rubocop-govuk/blob/main/config/rspec.yml#L23-L26) 
``` text
# We accept multiple expectations (within reason), preferring
# them to running mulitple similar tests.
RSpec/MultipleExpectations:
  Enabled: false
```

I don't see two expectations as a problem in this instance 🤷‍♂️

kmahern: Different versions of rubocop with different rules 🤯  . I think will leave as is in that case thanks both.

spec/services/calculators/income_contribution_calculator_spec.rb
kmahern: Thanks for the suggestion @jsugarman. I think I prefer to keep the it block for the description in this case and it is maybe more consistent with the other tests in this spec.

PR #868, opened 2022-04-01
Ap 3001 threshold changes

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3001)

Add new threshold file
Update tests to cover either side of the new thresholds
Added new tests to the CFE integration tests

## Checklist

### DIFF
--- a/spec/services/calculators/dependant_allowance_calculator_spec.rb
+++ b/spec/services/calculators/dependant_allowance_calculator_spec.rb
@@ -141,7 +141,7 @@ module Calculators
       end
     end
 
-    # Changing these to be 2021 dates.
+    # Changing these to be 2022 dates.
     describe "retrieving threshold values" do
       let(:dependant) { create :dependant }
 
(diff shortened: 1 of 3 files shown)


kmahern (approved):
lgtm 👍

spec/services/calculators/dependant_allowance_calculator_spec.rb
```diff
@@ -141,50 +141,50 @@ module Calculators
       end
     end
 
-    # Changing these to be 2021 dates.
```
colinbruce, line 144: It's hard to tell in the github view, but do we have a set of tests for each threshold file?  
I seem to remember this biting us when someone changed an old threshold file and the tests didn't fail.
I would suggest having a 2021 block, and add a 2022 block so we can test both thresholds

willc-work: no tests for each threshold file, just to straddle the most recent threhsold period. I will add tests for all dates

colinbruce: I wouldn't necessarily worry about _all_ dates, but I would definitely add new years. 
I seem to remember that someone did a search and replace on a value and changed multiple threshold files and we didn't spot it! 😬

willc-work: fixed

PR #899, opened 2022-04-29
El-171 - Create crime assessment

exonian:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/EL-171)

Adds the `assessment_type` to `Assessment` model, and allows creation of it via creator and controller
Adds an integration test for this behaviour on the controller
Adds a check to an existing test to ensure `assessment_type` remains unset if not passed

Partially handles not passing `ccms_codes`, but not completely. That needs some work, but the code in its current state is useful for building other behaviours (POSTs to other endpoints that attach objects to an Assessment) on

### DIFF
--- a/app/models/assessment.rb
+++ b/app/models/assessment.rb
@@ -48,6 +48,10 @@ class Assessment < ApplicationRecord
     version == "3"
   end
 
+  def criminal?
+    assessment_type == "criminal"
+  end
+
 private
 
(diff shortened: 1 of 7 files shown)


katylouise (approved):
I think this looks fine.  For now we can get away with using the dummy code for proceeding type as it should only have a bearing on capital (which we're not testing).

PR #917, opened 2022-05-12
[AP-3099] Add rswag documentation

jsugarman:
## What
Add interactive API documentation using rswag

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3099)

So we can interact with API via an GUI, and
remove need for jquery.

## Checklist


kmahern (approved):
lgtm 👍 @jsugarman although I think it needs a rebase? I wasn't able to take a look at the api-docs endpoint though as it didn't deploy to uat.

stephenrichards (commented):
Hi - the only comment I've got is that the default landing place for the swagger docs is V3, with V4 showing the POST assessement, which I think is the  only difference in the payloads.

We only use v4, and there is a plan to rip all the v3 stuff out.  I think it would be more useful just to document the v4.

jsugarman:
> Hi - the only comment I've got is that the default landing place for the swagger docs is V3, with V4 showing the POST assessement, which I think is the only difference in the payloads.
> 

👍  I have made V4 page default now and moved all endpoints to it

stephenrichards (approved):

PR #927, opened 2022-05-18
Ap 3113 remove apipie json schema

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-3113)

Removed APIPIE documentation and validation from applicants endpoint only.
The idea is this would then establish a pattern for the other endpoints.
Once the documentation and validation is removed from all endpoints the APIPE gem and config can be removed.
This approach uses json schema as a replacement for APIPIE for validating the request parameters.
The applicant schema is a .erb file to allow for the use of ref specifying an external file with common definitions.

## Checklist


app/validators/json_validator.rb
```diff
@@ -0,0 +1,22 @@
+class JsonValidator
+  def initialize(schema, payload)
+    @schema_dir = Rails.root.join("public/schemas")
```
stephenrichards, line 3: Just a slight niggle about naming: I think the incoming parameter and paramter passed to `#load_schema` should be renamed from `schema` to `schema_name` to avoid any confusion

app/controllers/applicants_controller.rb
```diff
@@ -48,4 +21,12 @@ def creation_service
...
     @input ||= JSON.parse(request.raw_post, symbolize_names: true)
   end
+
+  def json_validator
+    @json_validator ||= JsonValidator.new(schema, request.raw_post)
+  end
+
+  def schema
+    @schema ||= "applicant"
+  end
```
stephenrichards, line 31: I think `schema_name` should be used rather than `schema` to make it clear that it's just the name your're passing in and not the schema itself.

Also, I don't think there is any benefit in  momoizing the name of the schema.  The controller class will be instantiated for each incoming request, and just process one request, so it's a bit of an overkill. So I think lines 25 to 31 can simply be replaced with: 

```
def json_validator
  JsonValidator.new('applicant', request.raw_post)
end
```

public/schemas/applicant.json.erb
```diff
@@ -0,0 +1,28 @@
+{
+  "id": "http://localhost:3000/schemas/applicant.json",
```
stephenrichards, line 2: I think this should be:
```
"id": "<%= "file://#{@schema_dir}/applicant.json" %>",
```

public/schemas/applicant.json.erb
```diff
@@ -0,0 +1,28 @@
...
+  "schema": "http://json-schema.org/draft-04/schema#",
```
stephenrichards, line 3: I think the keyword here is `$schema`, not `schema`

public/schemas/applicant.json.erb
```diff
@@ -0,0 +1,28 @@
...
+    "applicant": {
+      "type": "object",
+      "required": ["date_of_birth", "involvement_type", "has_partner_opponent", "receives_qualifying_benefit"],
+      "properties": {
+        "date_of_birth": {
+          "$ref": "<%= "file://#{@schema_dir}/common.json#date" %>"
+        },
+        "involvement_type": {
+          "type": "string",
+          "pattern": "^applicant"
```
stephenrichards, line 17: a better way of doing this is:
```
"involvement_type": {
  enum: ["applicant"]
}
```

stephenrichards (changes requested):

app/validators/json_validator.rb
kmahern: Thanks @stephenrichards I have made the changes you requested.

app/validators/json_validator.rb
```diff
@@ -0,0 +1,22 @@
+class JsonValidator
+  def initialize(schema_name, payload)
+    @schema_dir = Rails.root.join("public/schemas")
+    @schema_name = load_schema(schema_name)
```
jsugarman, line 4: for clarity i would have a @schema_name and a @schema var. The @schema being the loaded :"schema"  of name @schema_name 

```
class JsonValidator
   def initialize(schema_name, payload)
     @schema_name = schema_name
     @schema = load_schema
   end
   
   
   private
   
   attr_reader :schema_name, :schema
   
   ...can amend to use attr_reader fpor schema_name and/or schema elsewhere then
```

Do you want to pair on this?!

spec/requests/applicants_controller_spec.rb
```diff
@@ -120,16 +120,27 @@
...
+      context "receives_qualifying_benefit not a boolean" do
+        it "fails with a message" do
           params[:applicant][:receives_qualifying_benefit] = "yes"
           post assessment_applicant_path(assessment.id), params: params.to_json, headers: headers
           expect(response).to have_http_status(:unprocessable_entity)
         end
       end
     end
+
+    context "future date of birth" do
```
jsugarman, line 135:
```suggestion
    context "with date of birth in the future" do
```

spec/requests/applicants_controller_spec.rb
```diff
@@ -120,16 +120,27 @@
         end
 
         it_behaves_like "it fails with message",
-                        %(Invalid parameter 'has_partner_opponent' value "yes": Must be one of: <code>true</code>, <code>false</code>, <code>1</code>, <code>0</code>.)
+                        /The property '#\/applicant\/has_partner_opponent' of type string did not match the following type: boolean in schema/
       end
 
-      context "for documentation" do
-        it "fails with a message", :show_in_doc do
+      context "receives_qualifying_benefit not a boolean" do
```
jsugarman, line 126:
```suggestion
      context "with non boolean receives_qualifying_benefit" do
```

spec/validators/json_validator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        applicant: {
+          date_of_birth: dob,
+          has_partner_opponent: opponent,
+        },
+      }.to_json
+    end
+
+    it "is not valid" do
+      expect(validator).not_to be_valid
+    end
```
jsugarman, line 57: RSpec supplies descriptions  anyway

```suggestion
    it {  expect(validator).not_to be_valid }
```

spec/validators/json_validator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    it "has the expected errors" do
+      errors = validator.errors
+      expect(errors).to include(match(/The property '#\/applicant' did not contain a required property of 'involvement_type' in schema/))
+      expect(errors).to include(match(/The property '#\/applicant' did not contain a required property of 'receives_qualifying_benefit' in schema/))
```
jsugarman, line 62: Could use a compound matcher
```suggestion
      expect(errors).to 
        include(match(/The property '#\/applicant' did not contain a required property of 'involvement_type' in schema/)).and 
        include(match(/The property '#\/applicant' did not contain a required property of 'receives_qualifying_benefit' in schema/))
```

app/validators/json_validator.rb
```diff
@@ -0,0 +1,22 @@
...
+  def errors
+    JSON::Validator.fully_validate(@schema_name, @payload)
+  end
+
+private
+
+  def load_schema(schema_name)
+    filename = "#{@schema_dir}/#{schema_name}.json.erb"
+    ERB.new(File.read(filename)).result(binding)
+  end
```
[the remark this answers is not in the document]
kmahern: Thanks @jsugarman have made the changes to json_validator we discussed

spec/validators/json_validator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+  end
+
+  context "when dob is invalid and involvement type not applicant" do
+    let(:dob) { "3002-12-23" }
+    let(:involvement_type) { "defendant" }
+
+    it "is not valid" do
+      expect(validator).not_to be_valid
+    end
```
jsugarman, line 36:
```suggestion
    it { expect(validator).not_to be_valid }
```

spec/validators/json_validator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  let(:benefit) { true }
+
+  let(:schema_name) { "applicant" }
+
+  let(:validator) { JsonValidator.new(schema_name, payload) }
+
+  context "when valid payload" do
+    it "returns true" do
+      expect(validator).to be_valid
+    end
```
jsugarman, line 27:
```suggestion
    it { expect(validator).to be_valid }
```

spec/validators/json_validator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+          has_partner_opponent: opponent,
+        },
+      }.to_json
+    end
+
+    it { expect(validator).not_to be_valid }
+
+    it "has the expected errors" do
+      errors = validator.errors
+      expect(errors)
```
jsugarman, line 60:
```suggestion
      expect(validator.errors)
```

spec/validators/json_validator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    let(:dob) { "3002-12-23" }
+    let(:involvement_type) { "defendant" }
+
+    it "is not valid" do
+      expect(validator).not_to be_valid
+    end
+
+    it "displays errors" do
+      errors = validator.errors
+      expect(errors)
```
jsugarman, line 40:
```suggestion
      expect( validator.errors)
```

jsugarman (approved):
A few more nits that i did not notice the first time but approving.

As discussed, we can look at how rswag could use the schemas to generate the documentation or schema could be generated from the rswag definitions to avoid repetition of, in particular, request body parameter definitions in rswag.

app/controllers/applicants_controller.rb
```diff
@@ -1,34 +1,7 @@
...
-    property :success, %w[true], desc: "Success flag shows true"
-    property :errors, [], desc: "Empty array of error messages"
-  end
-  returns code: :unprocessable_entity do
-    property :success, %w[false], desc: "Success flag shows false"
-    property :errors, array_of: String, desc: "Description of why object invalid"
-  end
-
   def create
+    return render_unprocessable(json_validator.errors) unless json_validator.valid?
```
jsugarman, line 3: ideally this should be internal to the creation service actually?

kmahern: @jsugarman I am not sure, I intentionally added return render_unprocessable(json_validator.errors) unless json_validator.valid? to stop it hitting the creation service if the validation failed (it previously didn’t hit the creation service if the apipie validation failed) - is that not a valid approach? (To me we are using json_schema to validate the request params so it makes sense to call it from the controller. We pass creation_service a hash input so we would probably have to turn this back into json in order to validate it with json_schema if we did that in the creation service). Happy to pair if you want to?

kmahern: @jsugarman I have added a commit to move the call to json_validator from the controller to the creator.

PR #945, opened 2022-06-01
Ap 3153 capitals controller to json

chris-groves:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-3153)

This is part of the project to remove APIPIE from CFE and replace it with RSWAG.  

APIPIE carries out a lot of validation of the incoming payload, which RSWAG does not, so we are going to use JSON schema to perform that validation.

This PR focuses on the Capitals Controller.

## Checklist

### DIFF
--- a/public/schemas/capital.json.erb
+++ b/public/schemas/capital.json.erb
@@ -1,0 +1,53 @@
+{
+  "id": "file://#{@schema_dir}/capital.json",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Legal Aid Check Financial Eligibility add capital details to an assessment schema",
+  "type": "object",
+  "properties": {
+    "bank_accounts": {
+      "description": "Describes the name of the bank account and the lowest balance during the computation period",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "description": {
+            "description": "The name of the bank account",
+            "type": "string"
+          },
+          "value": {
+            "$ref": "<%= "file://#{@schema_dir}/common.json#currency" %>"
+          }
+        },
+        "additionalProperties": false,
+        "required": [
+          "description",
+          "value"
+        ]
+      }
+    },
+    "non_liquid_capital": {
+      "description": "An array of objects describing applicant's non-liquid capital items (excluding property), e.g. valuable items, jewellery, trusts, other investments",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "description": {
+            "description": "Definition of a liquid or non-liquid capital item",
+            "type": "string"
+          },
+          "value": {
+            "description": "The value of the item",
+            "positive_currency": {
+              "$ref": "<%= "file://#{@schema_dir}/common.json#positive_currency" %>"
+            }
+          }
+        },
+        "additionalProperties": false,
+        "required": [
+          "description",
+          "value"
+        ]
+      }
+    }
+  }
+}
--- a/public/schemas/capital.json.erb
+++ b/public/schemas/capital.json.erb
@@ -36,10 +36,7 @@
             "type": "string"
           },
           "value": {
-            "description": "The value of the item",
-            "positive_currency": {
-              "$ref": "<%= "file://#{@schema_dir}/common.json#positive_currency" %>"
-            }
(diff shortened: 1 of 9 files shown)


public/schemas/capital.json.erb
```diff
@@ -0,0 +1,16 @@
+{
+  "id": "file://#{@schema_dir}/capital.json",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Legal Aid Check Financial Eligibility add capital details to an assessment schema",
+  "type": "object",
+  "properties": {
+    "bank_accounts": {
+      "$ref": "<%= "file://#{@schema_dir}/common.json#bank_accounts" %>"
```
kmahern, line 8: Are bank accounts or non_liquid_capital used in any other cfe endpoints? If so, I would define them here in capital.json.erb rather than in common.json (see https://github.com/ministryofjustice/check-financial-eligibility/blob/e3c395df416d07a939470970c111d2a0ff34a214/public/schemas/applicant.json.erb)

spec/requests/capitals_controller_spec.rb
```diff
@@ -98,25 +98,25 @@
...
 
-        it_behaves_like "it fails with message", "Missing parameter value"
+        it_behaves_like "it fails with message", "ActiveRecord::RecordInvalid: Validation failed: Value can't be blank"
       end
 
       context "missing description on non_liquid capital" do
         let(:non_liquid_params) { attributes_for_list(:non_liquid_capital_item, 2).map { |nlc| nlc.tap { |item| item.delete(:description) } } }
 
-        it_behaves_like "it fails with message", "Missing parameter description"
+        it_behaves_like "it fails with message", "ActiveRecord::RecordInvalid: Validation failed: Description can't be blank"
```
kmahern, line 113: It looks to me like the json-schema validation isn't working as expected if you are getting the ActiveRecord::RecordInvalid error. (To me this implies we are either not calling json_validator before trying to save the record or the validation is not correct).

spec/requests/swagger_docs/v4/capitals_spec.rb
```diff
@@ -104,7 +104,7 @@
 
         run_test! do |response|
           body = JSON.parse(response.body, symbolize_names: true)
-          expect(body[:errors]).to include(/Missing parameter value/)
+          expect(body[:errors]).to include("ActiveRecord::RecordInvalid: Validation failed: Value can't be blank")
```
kmahern, line 107: It looks to me like the json-schema validation isn't working as expected if you are getting the ActiveRecord::RecordInvalid error. (To me this implies we are either not calling json_validator before trying to save the record or the validation is not correct).

kmahern (changes requested):
@chris-groves to me it looks like we are relying on the active model validation you added on the capital_item model, and we are not seeing the json-schema validation working here (at least based on the changes to the specs). I would expect to see CFE returning something like The property '#\/' did not contain a required property of 'capital_item'/ (or similar messages).

kmahern (approved):
@chris-groves I have made a couple of suggested changes but I don't think they are blockers. I would also be tempted to add some additional tests in spec/requests/capitals_controller_spec.rb to test for invalid values for bank_account value and non_liquid_capital value (e.g. strings, negative values when should be positive etc.) - again probably not a blocker so approving.

public/schemas/capital.json.erb
```diff
@@ -0,0 +1,61 @@
...
+          "description": {
+            "description": "The name of the bank account",
+            "type": "string"
+          },
+          "value": {
+            "description": "The lowest balance on the account during the caluclation period",
+            "neg_or_pos_number_with_two_decimals": {
+              "description": "A negative or positive number (including zero) with two decimal places",
+              "type": "number",
+              "multipleOf": 0.01
```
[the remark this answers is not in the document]
kmahern: Apologies @chris-groves , I don't think this will work. I think the tests are failing because the tests are generating values in the format 0.44e5 (and this is meaning the regex pattern match fails). I think apply is also sending values in this format so I don’t think we can use the regex and will have to go back to using multipleOf - sorry!

colinbruce (approved):
LGTM :+1:

PR #967, opened 2022-06-16
Eligibility - adjusted income for crime assessments

katylouise:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/EL-169)

Added functionality to assess eligibility for criminal adjusted income.  This is to enable an initial means test for crime assessments.

## Checklist

### DIFF
--- a/app/services/creators/assessment_creator.rb
+++ b/app/services/creators/assessment_creator.rb
@@ -35,7 +35,7 @@ module Creators
         submission_date: Date.parse(@parsed_raw_post[:submission_date]),
         matter_proceeding_type: @parsed_raw_post[:matter_proceeding_type],
         proceeding_type_codes: ccms_codes_for_application,
-        assessment_type: assessment_type,
+        assessment_type:,
         version: @version,
         remote_ip:,
       }
--- a/app/services/creators/eligibilities_creator.rb
+++ b/app/services/creators/eligibilities_creator.rb
@@ -9,10 +9,14 @@ module Creators
     end
 
     def call
-      GrossIncomeEligibilityCreator.call(@assessment)
-      DisposableIncomeEligibilityCreator.call(@assessment)
-      CapitalEligibilityCreator.call(@assessment)
-      AssessmentEligibilityCreator.call(@assessment)
+      if @assessment.assessment_type == 'criminal'
+        AdjustedIncomeEligibilityCreator.call(@assessment)
+      else
+        GrossIncomeEligibilityCreator.call(@assessment)
+        DisposableIncomeEligibilityCreator.call(@assessment)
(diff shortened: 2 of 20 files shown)


app/services/creators/assessment_creator.rb
```diff
@@ -35,7 +35,7 @@ def generate_assessment_hash(remote_ip)
         submission_date: Date.parse(@parsed_raw_post[:submission_date]),
         matter_proceeding_type: @parsed_raw_post[:matter_proceeding_type],
         proceeding_type_codes: ccms_codes_for_application,
-        assessment_type: assessment_type,
+        assessment_type:,
```
exonian, line 38: Out of interest, what does this do?

app/services/creators/eligibilities_creator.rb
```diff
@@ -9,10 +9,14 @@ def initialize(assessment)
     end
 
     def call
-      GrossIncomeEligibilityCreator.call(@assessment)
-      DisposableIncomeEligibilityCreator.call(@assessment)
-      CapitalEligibilityCreator.call(@assessment)
-      AssessmentEligibilityCreator.call(@assessment)
+      if @assessment.assessment_type == "criminal"
```
exonian, line 12: I think this could be `if @assessment.criminal` or similar, I feel like that's slightly neater

app/services/creators/assessment_creator.rb
katylouise: I think this was rubocop (the Ruby linter) which made this change.  It is equivalent to saying `assessment_type: assessment_type`.

app/services/creators/eligibilities_creator.rb
katylouise: Agreed.  I can add this method and make this refactor as part of the assessment ticket I'm picking up next.

exonian: I think it's already there, but it's eminently possible I in fact put it in the wrong place...

exonian: I realise now the "or similar" made it look like I was speculating about what the method would be called, rather I just wasn't confident about the syntax of using it

app/services/creators/assessment_creator.rb
colinbruce: :wave: it was a new ruby rule.
If the key and value are the same, you can omit the value.
We discussed turning it off as it can make the code more awkward to read but it has stayed in for now.
```ruby
  {
    proceeding_type_codes: ccms_codes_for_application, # value explicitly passed
    assessment_type:, # value not needed because it matches the key name
  }
```

exonian: Ah just 🪄 things then 😅 
Thanks both!

PR #971, opened 2022-06-20
AP-3156 Employments controller to JSON Schema

chris-groves:
[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-3156)

This is part of the project to remove APIPIE from CFE and replace it with RSWAG.

APIPIE carries out a lot of validation of the incoming payload, which RSWAG does not, so we are going to use JSON schema to perform that validation.

This work focuses on the Employments Controller.

## Checklist

### DIFF
--- a/app/controllers/employments_controller.rb
+++ b/app/controllers/employments_controller.rb
@@ -1,40 +1,4 @@
 class EmploymentsController < ApplicationController
-  resource_description do
-    name "Employments"
-    short "Add details of an applicant's employment to an assessment"
-    formats(%w[json])
-    description <<-END_OF_TEXT
-    == Description
-      Adds details of an applicant's employment/earned income to an assessment.
-    END_OF_TEXT
-  end
-
-  api :POST, "assessments/:assessment_id/employments", "Create employment record"
-  formats(%w[json])
-  param :assessment_id, :uuid, required: true
-  param :employment_income, Array, of: Hash, desc: "Collection of employment and income financial information" do
-    param :name, String, required: true, desc: "An identifying name for this employment e.g. employer name"
-    param :client_id, String, required: true, desc: "An id supplied by the client to identify this employment"
-    param :payments, Array, of: Hash, required: true, desc: "A collection of information about income from the employment" do
-      param :client_id, String, required: true, desc: "An id supplied by the client to identify this payment"
-      param :date, Date, date_option: :today_or_older, required: true, desc: "The date payment received"
-      param :gross, :currency, currency_option: :not_negative, required: true, desc: "Gross income received figure"
-      param :benefits_in_kind, :currency, currency_option: :not_negative, required: true, desc: "Benefit in kind amount"
-      param :tax, :currency, required: true, desc: "Amount of tax paid"
-      param :national_insurance, :currency, required: true, desc: "Amount of national insurance paid"
-      param :net_employment_income, :currency, currency_option: :not_negative, required: true, desc: "Net income received figure"
-    end
-  end
-
-  returns code: :ok, desc: "Successful response" do
-    property :success, %w[true], desc: "Success flag shows true"
-    property :errors, [], desc: "Empty array of error messages"
-  end
-  returns code: :unprocessable_entity do
-    property :success, %w[false], desc: "Success flag shows false"
-    property :errors, array_of: String, desc: "Description of why object invalid"
-  end
-
   def create
     if creation_service.success?
       render_success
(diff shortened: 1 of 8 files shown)


colinbruce (approved):
LGTM :+1:

app/services/creators/employments_creator.rb
```diff
@@ -9,12 +9,20 @@ def initialize(assessment_id:, employments_attributes: [])
...
+      else
+        self.errors = json_validator.errors
+      end
       self
     end
 
   private
 
+    def json_validator
+      @json_validator ||= JsonValidator.new("employment", employments_incomes)
```
jsugarman, line 23: other PRs have been passing raw JSON params in here i think. This is passing a hash. surprised it is not causing a  causing a problem and perhaps we should be using a single pattern?

jsugarman (commented):
Just a question on pattern consistency with other similar PRs

app/controllers/employments_controller.rb
```diff
@@ -48,11 +12,7 @@ def create
   def creation_service
     @creation_service ||= Creators::EmploymentsCreator.call(
       assessment_id: params[:assessment_id],
-      employments_attributes: input[:employment_income],
+      employments_attributes: JSON.parse(request.raw_post, symbolize_names: true),
```
jsugarman, line 15: As discussed for parity with other controllers and creator services
```suggestion
      employments_params: request.raw_post
```

app/services/creators/employments_creator.rb
```diff
@@ -1,6 +1,6 @@
 module Creators
   class EmploymentsCreator < BaseCreator
-    attr_accessor :assessment_id, :employments_attributes
+    attr_accessor :assessment_id, :employments_incomes
```
jsugarman, line 3: do we need these? for parity

app/services/creators/employments_creator.rb
```diff
@@ -9,12 +9,20 @@ def initialize(assessment_id:, employments_attributes: [])
     end
 
     def call
-      create_records
+      if json_validator.valid?
+        create_records
+      else
+        self.errors = json_validator.errors
```
jsugarman, line 15: As discussed with KA and CB
```suggestion
        errors.concat(json_validator.errors)
```

app/services/creators/employments_creator.rb
```diff
@@ -1,6 +1,6 @@
 module Creators
   class EmploymentsCreator < BaseCreator
-    attr_accessor :assessment_id, :employments_attributes
+    attr_accessor :assessment_id, :employments_incomes
 
     def initialize(assessment_id:, employments_attributes: [])
```
jsugarman, line 5:
```suggestion
    def initialize(assessment_id:, employments_params: )
```

app/services/creators/employments_creator.rb
jsugarman: this is just the code that made me investigate further and discuss

jsugarman: use @employment_params here

app/services/creators/employments_creator.rb
```diff
@@ -25,7 +33,7 @@ def create_records
     end
 
     def create_employment
-      @employments_incomes.each do |employment|
+      @employments_incomes[:employment_income].each do |employment|
```
jsugarman, line 36: create a method

```
def employment_attributes
   @employment_attributes ||= JSON.parse(...blah, symbolize: true)
end
```
and use above
end
```

File: public/schemas/employment.json.erb [hunk 7]
```diff
@@ -0,0 +1,67 @@
+{
+  "id": "file://#{@schema_dir}/employment.json.erb",
```
jsugarman, line 2:
```suggestion
  "id": "file://public/schema/employment.json.erb",
```
or dynamic

app/services/creators/employments_creator.rb
```diff
@@ -47,7 +57,7 @@ def create_payments(employment)
     end
 
     def assessment
-      @assessment ||= Assessment.find_by(id: assessment_id) || (raise CreationError, ["No such assessment id"])
+      @assessment ||= Assessment.find_by(id: @assessment_id) || (raise CreationError, ["No such assessment id"])
```
jsugarman, line 60: **not a blocker**, This is called at line 30 to raise errors but is not subsequently used. It could be used in place of @assesment_id (use assessment.id)

jsugarman (approved):
👍 ✅  great, one non-blocking comment

PR #980, opened 2022-06-22
Remove APIPIE from other_incomes controller

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3159)

## Checklist

### DIFF
--- a/app/services/creators/other_incomes_creator.rb
+++ b/app/services/creators/other_incomes_creator.rb
@@ -6,15 +6,19 @@ module Creators
 
     attr_reader :other_income_sources
 
-    def initialize(assessment_id:, other_incomes: [])
+    def initialize(assessment_id:, other_incomes_params:)
       super()
       @assessment_id = assessment_id
-      @other_incomes = other_incomes[:other_incomes]
+      @other_incomes_params = other_incomes_params
       @other_income_sources = []
     end
 
     def call
-      create_records
+      if json_validator.valid?
+        create_records
+      else
+        self.errors = json_validator.errors
+      end
       self
     end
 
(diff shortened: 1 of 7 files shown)


app/services/creators/other_incomes_creator.rb
```diff
@@ -56,5 +60,13 @@ def assessment
     def normalize(name)
       name.underscore.tr(" ", "_")
     end
+
+    def other_incomes
+      @other_incomes ||= JSON.parse(@other_incomes_params, symbolize_names: true)[:other_incomes]
```
jsugarman, line 65: **nit**, for rubocop future proofing, and my preferred mechanism
```suggestion
      @other_incomes ||= JSON.parse(@other_incomes_params, symbolize_names: true).fetch(:other_incomes, nil)
```

app/services/creators/other_incomes_creator.rb
```diff
@@ -6,15 +6,19 @@ class OtherIncomesCreator < BaseCreator
...
+      @other_incomes_params = other_incomes_params
       @other_income_sources = []
     end
 
     def call
-      create_records
+      if json_validator.valid?
+        create_records
+      else
+        self.errors = json_validator.errors
```
jsugarman, line 20: not a blocker/for discussion I have changed this in my APIPIE to below as errors method is instantiated to an empty array and ideally errors should be pushed/added to it rather than overwriting the entire array (which would make the code more breakable if something were to add an error first then this overwrites)
```suggestion
         errors.concat(json_validator.errors)
```

jsugarman (approved):
couple of nits/suggestions. no blockers

app/services/creators/other_incomes_creator.rb
kmahern: Thanks 👍 have made suggested change

app/services/creators/other_incomes_creator.rb
kmahern: Thanks 👍 have made suggested change

PR #990, opened 2022-06-28
Update README

chris-groves:
Update README with details of migration of API documentation from APIPIE to Rswag.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,12 +14,11 @@ It's defined as code and [can be edited](https://github.com/ministryofjustice/la
 
 ## Documentation
 
-The API is documented by APIPIE at `/apidocs`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -153,7 +152,7 @@ options:
 -w XXX    Only process worksheet named XXX
 ```
 
-Each worksheet has a entry `Test Active` which can be either true or false.  If set to false, the worksheet will be skipped, unless it is 
+Each worksheet has a entry `Test Active` which can be either true or false.  If set to false, the worksheet will be skipped, unless it is
```
jsugarman, line 155:
```suggestion
Each worksheet has an entry `Test Active` which can be either true or false.  If set to false, the worksheet will be skipped, unless it is
```

PR #1005, opened 2022-07-13
AP-3297 Calculation to take Client Involvement Type into consideration

stephenrichards:
## Calculation to take Client Involvement Type into consideration

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3297)

Up until now, client involvement type has not been considered in CFE, because Apply only allowed one type: Applicant.  Now that multiple client involvement types are allowed, it has a bearing on whether or not the upper thresholds are waived or not.  In short, for Client involvement type "A" (applicant) upper thresholds are waived for Domestic Abuse cases only.  In all other cases, they remain as specified by the thresholds.

In accordance with the principle of only one source for information, the Legal Framework API (LFA) is the source for this information, and so CFE needs to ask LFA whether or not to waive the thresholds for each proceeding type / client involvement type combination.

In a previous  PR, a Proceeding Types record was created, and this is written when Apply requests an assessment with version 5, and sends proceeding type/client information data to the new `/proceeding_types` endpoint.

This change does the following to enable the calculations to take client involvement type into consideration:
- adds three fields to the ProceedingType model:
  - `capital_upper_threshold`
  - `gross_income_upper_threshold`
  - `disposable_income_upper_threshold`
- At the start of the main workflow, a new service is called for version 5 assessments which queries LFA with all the proceeding type/client involvement type combinations on the assessment, and then populates these three fields with eithet the standard upper thresholds or the normal upper thresholds as appropriate
- later in the workflow, when the Eligibility records are created, for version 5 assessments, these thresholds are copied from the ProceedingType (for earlier versions, the existing method is still. used until versions 3 and 4 are deprecated)

## Checklist


app/services/creators/gross_income_eligibility_creator.rb
```diff
@@ -10,7 +10,13 @@ def initialize(assessment)
...
 
     def call
-      @assessment.proceeding_type_codes.each { |ptc| create_eligibility(ptc) }
+      # TODO: remove if statement once version 4 deprecated
+      if @assessment.version_5?
+        @assessment.proceeding_types.map(&:ccms_code).each { |ptc| create_eligibility(ptc) }
+      else
+        @assessment.proceeding_type_codes.each { |ptc| create_eligibility(ptc) }
+      end
+      # @assessment.proceeding_type_codes.each { |ptc| create_eligibility(ptc) }
```
kmahern, line 19: remove duplicated/commented out code.

kmahern (approved):
One small comment otherwise lgtm.

PR #1010, opened 2022-07-14
El 170 crime workflow

katylouise:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/EL-170)

Implement a new Workflow, Assessors and Decorators to allow a crime assessment to be requested and calculated.

## Checklist


exonian (approved):
I don't have much to say on the details of the code itself.

Reading through it, it definitely reinforces what we discussed: there's an awful lot of switches in this, and that could be improved somewhat by having more specialised classes with consistent interfaces. 
So that instances of them would be attached eg to `Assessment` at instantiation ie you'd have `assessment.workflow`, `assessment.decorator` and then `Assessment` would just always call the same things after init and the behaviour would depend on the classes and states of the models attached to it.
But also as discussed, I think we will need to go further than that and have the core CFE library act as a framework but have all configuration of models, services etc be done by the civil and crime consumers ie it would be good to build a crime assessment, a crime applicant and a crime decorator from common CFE components, and have them used by common CFE machinery, but not actually have a runtime where these crime components co-exist with either civil or even core components: the core ones would be used as building bricks only.

I think this PR in this current state is useful, for showing how complex it would be if we just added Crime equivalents alongside Civil ones without deliberately moving code up into a core layer.

PR #1041, opened 2022-08-10
Update setup instructions with notes TAP have made along the way

forceedge01:
## What

TAP has gone through the setup as per the readme and found a few bits that we're missing. We've added these notes in to the readme.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,9 @@ ALLOW_FUTURE_SUBMISSION_DATE
 ```
 
 Set ALLOW_FUTURE_SUBMISSION_DATE to true to allow integration tests to run with submission dates that are in the future.
-A copy of the `.env` file including the current values can be found in the `Shared-LAA` section of LastPass.
+A copy of the `.env` file including the current values can be found in the `Shared-LAA` section of LastPass (reach out to the team if you don't have access).
+
+Note: Remove “GOOGLE_SHEETS_” from several of the variables in .env file
 
 ## Threshold configuration files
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,7 +100,9 @@ ALLOW_FUTURE_SUBMISSION_DATE
 ```
 
 Set ALLOW_FUTURE_SUBMISSION_DATE to true to allow integration tests to run with submission dates that are in the future.
-A copy of the `.env` file including the current values can be found in the `Shared-LAA` section of LastPass.
+A copy of the `.env` file including the current values can be found in the `Shared-LAA` section of LastPass (reach out to the team if you don't have access).
+
+Note: Remove “GOOGLE_SHEETS_” from several of the variables in .env file
```
colinbruce, line 105: Did you download the correct file?  I cannot see any `GOOGLE_SHEETS_` lines in the CFE .env file.
There _are_ some in the *Apply* file though 😕

README.md
```diff
@@ -116,6 +118,11 @@ for testing new thresholds before they come into affect on production.
 
 ## Setup
 
+Please install the following dependencies prior to running the application setup:
+
+brew install shared-mime-info
+brew install postgresql
```
colinbruce, line 124: Fair point - most developers would have installed these as Apply which is generally the first service we use :+1:

colinbruce (commented):
Looks good, but the `GOOGLE_SHEETS_` comment may be on the wrong file

naseberry:
Author left org.
PR replaced by https://github.com/ministryofjustice/check-financial-eligibility/pull/1233

PR #1047, opened 2022-08-15
AP-3340 CFE: Remove legacy code for V3, V4. 

stephenrichards:
## Remove all legacy code for V3 and V4

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3340)

The only valid version of the CFE API is now V5, but there is a lot of code int he repo that handles and tests earlier versions (v3 and v4).  This PR removes it.

- removed the columns `matter_proceeding_type` and `proceeding_type_codes` from `assessments` table (this info is now held in a separate `proceeding_types` table)
- wrote `#proceeding_type_codes` method on `Assessment` to grab the codes from the `proceeding_types` table (previously this would have just come from the `proceeding_type_codes` column)
- Added the `proceedings` transient in the assessment factory to enable assessments to be written with specified proceeding types and client involvement types, and changed specs to use that instead of creating proceeding type records manually
- Most V5 decorators were identical to V3/V4 decorators - just aliases of the earlier versions.  All of these have been moved to the V5 namespace.
- All code where there was branches depending on the version has been eliminated - only the V5 path is preformed
- Default and valid versions in `CFEConstants` has been updated to be V5 only.
- Two migration helpers used to migrate to V4 and then V5 have been deleted (`EligibilityPopulator` and `ProceedingTypeCodeSeeder`)
- Updated the JSON schema for assessment to remove the `proceeding_matter_type` and `proceeding_types` attributes

## Checklist


colinbruce (commented):
Code-wise looks good, but we should fix the deploy before merging!

colinbruce (dismissed):

colinbruce (approved):

PR #1075, opened 2022-09-02
Change verbosity of integration tests

jsugarman:
## What
These are running silently and no output is produced in the github action 
for the BA to see.

Seems to be because the VERBOSE env var is set to true which translates
to 0 (silent verbosity) in `spec/integration/test_runner_spec.rb`.

## Checklist

### DIFF
--- a/.github/workflows/manual-integration-tests.yml
+++ b/.github/workflows/manual-integration-tests.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Run Tests
         env:
           SKIP_COVERAGE: true
-          VERBOSE: true
+          VERBOSE: 1
           CLIENT_EMAIL: ${{ secrets.CLIENT_EMAIL }}
           CLIENT_ID: ${{ secrets.CLIENT_ID }}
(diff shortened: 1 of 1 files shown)


cpjmcquillan (approved):

jsugarman:
@colinbruce when you are back. 

Not sure the silent running of the integration tests via the github action was was intentional or not. I suspect not as, 6 months ago it was set back to `ENV["VERBOSE"].to_i`, which is 0 when VERBOSE is "true", hence silent running. Dave does not need to the results soon anyway so can wait.

colinbruce:
> @colinbruce when you are back.
> 
> Not sure the silent running of the integration tests via the github action was was intentional or not. I suspect not as, 6 months ago it was set back to `ENV["VERBOSE"].to_i`, which is 0 when VERBOSE is "true", hence silent running. Dave does not need to the results soon anyway so can wait.

Indeed, I think when I first wrote the integration it was a boolean choice.  Then it got expanded to the current enum.  
Happy for this to be merged 👍

PR #1076, opened 2022-09-02
[AP-3439] add regular transactions

jsugarman:
## What
Add regular transactions model and include in gross income and disposable income summaries

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3439)

- Create regular_transactions table
- Add regular_transactions model, validations, factory and specs
- Add regular_transactions creator service and spec
- Add regular_transactions route, controller and spec
- Add regular_transactions swagger docs
- Add regular_income_collator for use in aggregating the `credit` regular_transactions onto the `gross_income_summary`
- Add regular_outgoings_collator for use in aggregating the `debit` regular_transactions onto the `disposable_income_summary`
- Amend to `HouseCostCalculator` to include regular_transactions for `rent_or_mortgage`
- Add integration spreadsheet and test case objects for regular transactions

This is the first stage of being able to produce an eligibility result for the How much How often flow
in civil apply. see [doc](https://docs.google.com/document/d/12ZyH6a6umqDrlB0FXrIuvJiX__UTtNXX1hXWiTjYe2k/edit#)  for eventual intentions.

## Checklist

### DIFF
--- a/app/controllers/regular_transactions_controller.rb
+++ b/app/controllers/regular_transactions_controller.rb
@@ -1,0 +1,18 @@
+class RegularTransactionsController < ApplicationController
+  def create
+    if creation_service.success?
+      render_success
+    else
+      render_unprocessable(creation_service.errors)
+    end
+  end
+
+private
+
+  def creation_service
+    @creation_service ||= Creators::RegularTransactionsCreator.call(
+      assessment_id: params[:assessment_id],
+      regular_transaction_params: request.raw_post,
+    )
+  end
+end
--- a/app/services/collators/disposable_income_collator.rb
+++ b/app/services/collators/disposable_income_collator.rb
@@ -4,8 +4,6 @@ module Collators
 
     attr_accessor :monthly_cash_transactions_total
 
-    OUTGOING_CATEGORIES = CFEConstants::VALID_OUTGOING_CATEGORIES.map(&:to_sym)
-
     delegate :net_housing_costs,
              :rent_or_mortgage_bank,
              :rent_or_mortgage_cash,
@@ -31,10 +29,14 @@ module Collators
 
   private
 
+    def outgoing_categories
+      CFEConstants::VALID_OUTGOING_CATEGORIES.map(&:to_sym)
+    end
+
     def populate_attrs
       attrs = {}
 
-      OUTGOING_CATEGORIES.each do |category|
+      outgoing_categories.each do |category|
         monthly_cash_amount = category == :child_care ? __send__("#{category}_cash") : monthly_cash_by_category(category)
         @monthly_cash_transactions_total += monthly_cash_amount unless category == :rent_or_mortgage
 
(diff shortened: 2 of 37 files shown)


app/services/collators/disposable_income_collator.rb
```diff
@@ -57,17 +61,25 @@ def default_attrs
...
-      + @monthly_cash_transactions_total - fixed_employment_allowance - employment_income_deductions
+      net_housing_costs +
+        dependant_allowance +
+        child_care_bank +
+        maintenance_out_bank +
+        legal_aid_bank +
+        @monthly_cash_transactions_total -
+        fixed_employment_allowance -
+        employment_income_deductions
     end
```
stephenrichards, line 83: I think we need to make it a bit clearer what is going on  here - it seems to be taking only the bank transactions of child_care, maintenance_out and legal _aid, but using the summed total in @monthly cash_transactions as well.

Perhaps we'd be better off writing a method `monthly_bank_transactions` that adds those three together, and then just adding those in this method

stephenrichards (changes requested):

stephenrichards (approved):

app/controllers/regular_transactions_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+  end
+
+private
+
+  def creation_service
+    @creation_service ||= Creators::RegularTransactionsCreator.call(
+      assessment_id: params[:assessment_id],
+      regular_transaction_params: request.raw_post,
+    )
```
cpjmcquillan, line 16: Just wondering if we need to memo-ize this?

It looks like it's only called once per request-response cycle, so I'd be inclined not to.

cpjmcquillan (approved):
The way we're collating different transactions feels quite brittle, and it doesn't feel particularly easy to follow.

I'm conscious I've never looked at the codebase before though, and the domain is unfamiliar to me.

Without delving into the code I don't feel I can make a strong recommendation or suggest areas for refactor, so I'm tempted to leave it for now (considering there seems to be a good range of tests). If this is something we agree we'd like to improve, let's come back to it at a later date (and add a ticket to the backlog).

I left some small comments, but you can take them or leave them.

Nice work making sense of it all!

🏆

app/controllers/regular_transactions_controller.rb
jsugarman: ty, this just follows the pattern of existing controllers and is quite probably redundant but would be inclined to leave unless we apply to all

db/migrate/20220901175043_create_regular_transactions.rb
```diff
@@ -0,0 +1,12 @@
+class CreateRegularTransactions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :regular_transactions, id: :uuid do |t|
+      t.belongs_to :gross_income_summary, foreign_key: true, null: false, type: :uuid
```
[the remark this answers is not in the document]
jsugarman: Personally I feel the consistency with other existing tables is significant  nd these are all using the  model/ORM dependant: destroy mechanism, but I also, more generally, rather like the flexibility of enforcement via the model/ORM rather than the database (which i have seen cause issues in the vague distant past)

PR #1087, opened 2022-09-13
Remove apipie

kmahern:
## What

This PR removes api-pie from the CFE codebase (including the apipie validators, which are no longer used).
It also removes some redundant code which was causing the pr to fail on coverage
It adds require active_support/parameter_filter to the sentry initialiser, as this was causing deployment to fail.

## Checklist


naseberry (approved):
💯

cpjmcquillan (approved):
😍 what a fabulous diff!

PR #1091, opened 2022-09-14
[AP-3474] Enable 3 monthly frequency for regular transactions

jsugarman:
## What
Add three_monthly equivalent frequency

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3474)

This is required as a design option for
presentation in the Apply service. There
therefore should be a matching frequency in CFE
for consistency and to avoid calculations on
the Apply side.

## Checklist

### DIFF
--- a/app/services/concerns/monthly_equivalent_calculatable.rb
+++ b/app/services/concerns/monthly_equivalent_calculatable.rb
@@ -1,21 +1,14 @@
 module MonthlyEquivalentCalculatable
 private
 
-  def determine_calc_method(period)
-    case period.to_sym
-    when :monthly
-      :monthly_to_monthly
-    when :four_weekly
-      :four_weekly_to_monthly
-    when :two_weekly
-      :two_weekly_to_monthly
-    when :weekly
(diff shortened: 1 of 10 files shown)


app/services/concerns/monthly_equivalent_calculatable.rb
```diff
@@ -1,21 +1,14 @@
...
-      raise ArgumentError, "unexpected period #{period}"
-    end
+  def determine_calc_method(frequency)
+    raise ArgumentError, "unexpected frequency #{frequency}" unless frequency_conversions.key?(frequency)
+
+    frequency_conversions[frequency]
+  end
+
+  def three_monthly_to_monthly(value)
+    (value / 3).round(2)
```
cpjmcquillan, line 11: I do find it concerning that we're doing floating point arithmetic with monetary values —particularly given the potential negative impact a variation could have on someone's eligibility result.

That is probably worth a wider discussion though, because it's prevalent throughout several of the services.

cpjmcquillan (approved):
:shipit:

PR #1095, opened 2022-09-15
BUG FIX - valdiation rules for Assessment

stephenrichards:
## BUG FIX - valdiation rules for Assessment

After the removal of code to deal with versions 3 and 4, there was a bug wherby proceeding_type_codes were 
being validated for version 5 (they were only needed for version 4).

This is a quick change to fix the live bug - but all the code relating to version 3 and 4 and special validation needs to be removed from Assessment and will be done in a subsquent ticket.

This also removes the spec file to test a particular version 3 bug which should have been removed as part of the merge of AP-3340

## Checklist


jsugarman (approved):
👍

PR #1099, opened 2022-09-16
Fix: Allow the test action to be triggered via Github

colinbruce:
## What

This has been triggered remotely via the slack bot, but that will be turned off for a while.  This alternative should allow users to trigger it via the actions page on github and reduce the need for Devs to run tests for non-dev staff

## Checklist


jsugarman (approved):
read [this on workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) and LGTM 👍

PR #1105, opened 2022-09-20
Refactor so that it doesn't appear assign the value to itself

stephenrichards:
## Refactor to make code more readable

The line `attrs[:"#{category}_bank"] = __send__("#{category}_bank")` was confusing because it appeared as if you were just adding the value of the column to a hash which would then be re-assigned back to the column.

This has been removed, and the subsequent line which relied on it changed to take the value from the column.

## Checklist

### DIFF
--- a/app/services/collators/disposable_income_collator.rb
+++ b/app/services/collators/disposable_income_collator.rb
@@ -39,22 +39,28 @@ module Collators
       attrs = {}
 
       outgoing_categories.each do |category|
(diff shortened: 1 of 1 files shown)


jsugarman (dismissed):
LGTM 👍 but have left possible further refactor suggestion

cpjmcquillan (approved):

colinbruce:
Closing as this was being done to make the partner transition easier and, with the subsequent CFE Civil cutover, it now seems a bit superfluous

PR #1109, opened 2022-09-21
AP-3496 Return error payload instead of raising an exception

stephenrichards:
## Return error payload rather than raising exception if assesment is requested before applicant and proceeding types have been added

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3946)

## Checklist

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -11,13 +11,11 @@ class AssessmentsController < ApplicationController
     if assessment_incomplete?
       render json: Decorators::ErrorDecorator.new(incomplete_message).as_json, status: :unprocessable_entity
     else
-      begin
-        perform_assessment
-      rescue StandardError => e
-        Sentry.capture_exception(e)
-        render json: Decorators::ErrorDecorator.new(e).as_json, status: :unprocessable_entity
-      end
+      perform_assessment
     end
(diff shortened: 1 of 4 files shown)


app/controllers/assessments_controller.rb
```diff
@@ -8,14 +8,28 @@ def create
...
+    if assessment_incomplete?
+      render json: Decorators::ErrorDecorator.new(incomplete_message).as_json, status: :unprocessable_entity
+    else
+      begin
+        perform_assessment
+      rescue StandardError => e
+        Sentry.capture_exception(e)
+        render json: Decorators::ErrorDecorator.new(e).as_json, status: :unprocessable_entity
+      end
+    end
```
jsugarman, line 20: **not a blocker** could move rescue out of if/else statement or even to a rescue_from, but maybe latter is too general?!

```suggestion
    if assessment_incomplete?
      render json: Decorators::ErrorDecorator.new(incomplete_message).as_json, status: :unprocessable_entity
    else
        perform_assessment
     end
   rescue StandardError => e
      Sentry.capture_exception(e)
      render json: Decorators::ErrorDecorator.new(e).as_json, status: :unprocessable_entity
    end
```

```
```

jsugarman (approved):
LGTM - couple of nits/suggestions but either way 👍

PR #1111, opened 2022-09-22
AP-3313 Cronjob to delete stale assessments

stephenrichards:
## Cronjob to delete stale assessments

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3313)

## Checklist


app/services/cull_stale_assessments_service.rb
```diff
@@ -0,0 +1,10 @@
+class CullStaleAssessmentsService
+  def self.call
+    new.call
+  end
+
+  def call
+    assessments = Assessment.where(created_at: 10.years.ago..2.weeks.ago)
+    assessments.map(&:destroy!)
```
jsugarman, line 8: would `destroy_all` not work here?

```suggestion
    assessments.destroy_all
```

Or even

```suggestion
   Assessment.destroy_by(created_at: 10.years.ago..2.weeks.ago)
```

jsugarman (approved):
LGTM 👍 - couple of suggestions.

❓ is keeping all but last two weeks meeting a particular need? Just thinking about if we wanted to investigate a historic request.

app/services/cull_stale_assessments_service.rb
jsugarman: You can also use infinite ranges i believe.
```suggestion
  Assessment.destroy_by(created_at: ..2.weeks.ago)
```

app/services/cull_stale_assessments_service.rb
```diff
@@ -4,6 +4,6 @@ def self.call
   end
 
   def call
-    Assessment.destroy_by(created_at: ..2.weeks.ago)
+    Assessment.destroy_by(created_at: ..CFEConstants::STALE_ASSESSMENT_THRESHOLD_DAYS.days.ago)
```
jsugarman, line 7: 👍 cool!

PR #1120, opened 2022-09-27
AP-3509 Gracefully prevent calling GET on the same assessment twice

stephenrichards:
## Gracefully prevent calling GET on the same assessment twice

Users reported getting exceptions when calling GET more than once for the same assessment.

This was because during the assessment process, eligibilty records are created and were failing because they already existed.  In order to enable calling GET on the assessment to be truly idempotent, we wold have had to delete any eligibilty records, remarks records and zeroise all the totals on the summary records.

For that reason it was felt better to gracefully refuse to carry out a seconds assesment as this would not be a normal use-case anyay, and return and explanatory message.

The swagger documentation has been updated to reflect this.

The error being raised with this was:
```
PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "eligibilities_unique_type_ptc" 
...
```

## Checklist

### DIFF
--- a/app/controllers/assessments_controller.rb
+++ b/app/controllers/assessments_controller.rb
@@ -8,8 +8,11 @@ class AssessmentsController < ApplicationController
   end
 
   def show
-    if assessment_incomplete?
+    case analyse_assessment_status
+    when :incomplete
       render json: Decorators::ErrorDecorator.new(incomplete_message).as_json, status: :unprocessable_entity
+    when :not_pending
+      render json: Decorators::ErrorDecorator.new(not_pending_message).as_json, status: :unprocessable_entity
     else
       perform_assessment
     end
@@ -20,12 +23,24 @@ class AssessmentsController < ApplicationController
 
 private
 
+  def analyse_assessment_status
+    return :incomplete if assessment_incomplete?
+
+    return :pending if assessment.assessment_result == "pending"
+
+    :not_pending
+  end
+
   def assessment_incomplete?
     assessment.proceeding_types.empty? || assessment.applicant.nil?
   end
 
   def incomplete_message
     "You must add proceeding types and applicant to before calling for the assessment to be calculated"
+  end
+
+  def not_pending_message
+    "Unable to call GET on the same assessment twice - please create new assessment and resubmit data"
   end
 
   def perform_assessment
--- a/spec/requests/swagger_docs/v5/assessments_spec.rb
+++ b/spec/requests/swagger_docs/v5/assessments_spec.rb
@@ -67,6 +67,11 @@ RSpec.describe "V5 Assessments", type: :request, vcr: true, swagger_doc: "v5/swa
       tags "Assessment"
       produces "application/json"
 
+      description <<~DESCIPTION.chomp
+        This endpoint will cause an assessment to be calculated using all the data that has been uploaded to the
+        other endpoints and return the result.  You call this endpoint twice for the same assessment_id.
+      DESCIPTION
+
       response(200, "successful") do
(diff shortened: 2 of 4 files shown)


app/controllers/assessments_controller.rb
```diff
@@ -8,8 +8,11 @@ def create
   end
 
   def show
-    if assessment_incomplete?
+    case analyse_assessment_status
+    when :incomplete
       render json: Decorators::ErrorDecorator.new(incomplete_message).as_json, status: :unprocessable_entity
+    when :not_pending
```
jsugarman, line 14: would `:already_assessed` or similar be clearer?

spec/requests/swagger_docs/v5/assessments_spec.rb
```diff
@@ -67,6 +67,11 @@
       tags "Assessment"
       produces "application/json"
 
+      description <<~DESCIPTION.chomp
+        This endpoint will cause an assessment to be calculated using all the data that has been uploaded to the
+        other endpoints and return the result.  You call this endpoint twice for the same assessment_id.
```
jsugarman, line 72: should this be 
```suggestion
        other endpoints and return the result.  You cannot call this endpoint twice for the same assessment_id.
```

swagger/v5/swagger.yaml
```diff
@@ -104,6 +104,9 @@ paths:
       summary: show assessment
       tags:
       - Assessment
+      description: |-
+        This endpoint will cause an assessment to be calculated using all the data that has been uploaded to the
+        other endpoints and return the result.  You call this endpoint twice for the same assessment_id.
```
jsugarman, line 109: [ditto](https://github.com/ministryofjustice/check-financial-eligibility/pull/1120/files#r981402693)

jsugarman (approved):
LGTM 👍

colinbruce:
Is it worth having a chat with the EFE team about their requirements?   I believe they were originally looking at a DB-less system that may need to render our info twice and, therefore, GET the assessment twice.

Rather than provide a nice error, would simply return the assessment record if it exists, or create it first if it doesn't make more sense?

jsugarman:
Sure, I think a GET should be repeatable but I don't understand what the implications are for say retrieving the assessment a few days later. is there some date sensitivity?! In addition if I GET the assessment, then add some more outgoings or other data, does the GET need to do some kind of update as Stephen alludes to...

> In order to enable calling GET on the assessment to be truly idempotent, we wold have had to delete any eligibilty records, remarks records and zeroise all the totals on the summary records.

Or do we need to prevent adding data on other endpoints once the GET has been performed for a given assessment?!

cpjmcquillan:
_disclaimer: I haven't look at any of the code, and I'm going entirely off what is being described as the system's behaviour_

If a GET request is doing anything other than retrieving some data, should it be a GET request at all?

It sounds like we're in a bit of a pickle and the API conflates _doing_ an assessment with _retrieving_ an assessment result.

colinbruce:
> I don't want to block this change if the error is impacting or blocking another team, but I do think this is something that is important and we should resolve —ideally before the API is used more widely.

I think taking our time is okay - EFE raised a query in Slack about why they couldn't call the endpoint twice and I added a backlog ticket to discuss and identify a solution.

Stephen picked it up and proposed this... I think we should probably take our time and discuss as a team with EFE to take a more measured approach to a solution

colinbruce:
Marked as OnHold pending an Apply/EFE Lead Dev meeting to discuss ways of working/best way of implementing this

colinbruce:
Closing as partner/civil have come up with a one-shot api that negates the need for this

PR #1140, opened 2022-10-18
Allow 'unspecified_source_income'/'quarterly' irregular payment type

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/EL-400)

We want to capture income that is from an unspecified source, expressed by the user as a single total figure received over the past 3 months. By 'unspecified source' we mean income that doesn't fall under any of the named types of regular or irregular payments that CFE accepts, with no requirement on the user to describe the source in any way.

So this PR adds 'unspecified_source_income' as a valid irregular type, and 'quarterly' as a valid income frequency, to the irregular payments model/endpoint.

Also, because prior to this there was only one irregular payment type - student loans - there was a bit of logic that assumed _all_ irregular payments were student loans, an assumption that started to break with the new type. So I unpicked that logic.

### DIFF
--- a/app/constants/cfe_constants.rb
+++ b/app/constants/cfe_constants.rb
@@ -57,9 +57,11 @@ module CFEConstants
   # Irregular income categories and frequencies
   #
   ANNUAL_FREQUENCY = "annual".freeze
+  QUARTERLY_FREQUENCY = "quarterly".freeze
   STUDENT_LOAN = "student_loan".freeze
-  VALID_IRREGULAR_INCOME_FREQUENCIES = [ANNUAL_FREQUENCY].freeze
-  VALID_IRREGULAR_INCOME_TYPES = [STUDENT_LOAN].freeze
+  UNSPECIFIED_SOURCE_INCOME = "unspecified_source_income".freeze
+  VALID_IRREGULAR_INCOME_FREQUENCIES = [ANNUAL_FREQUENCY, QUARTERLY_FREQUENCY].freeze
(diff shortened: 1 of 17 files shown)


colinbruce:
@jsugarman can you take a look at this?
I've run the tests locally and see no issue with our scenarios, but I know 🟡 is doing more transaction changes so this may impact your team more?

jsugarman:
have started to review this @colinbruce @patrick-laa  - will finish off thoughts and publish asap. 

Am wondering on first glance whether a new category of unspecified regular transaction (operation credit, frequency three_monthly) might do?

patrick-laa:
Thanks for agreeing to review this, @jsugarman! The reason we've proposed treating this as an 'irregular' income is that, as I understand it, we're expecting users to use the field to report in aggregate sums of money that are not received at fixed intervals in fixed amounts, rather than regular recurring payments. But I appreciate that within the context of CFE, 'regular' and 'irregular' may have more specific meanings than that. May I ask if there's a reason why treating this sort of thing as 'regular' payments within CFE would be preferable?

jsugarman:
> May I ask if there's a reason why treating this sort of thing as 'regular' payments within CFE would be preferable?

Merely my first thoughts on seeing the "quarterly" frequency being added because we had only recently put in the similar idea of regular income or outgoings for a "quarter" via the regular transaction model. 

I am not very familiar with the logic and codebase of CFE so would need to check the differences in handling of irregular and regular before commenting further. We have also discussed changing permissions on this repo as, for one thing, the test don't run when you open a PR.

PR #1150, opened 2022-10-24
Validate date_of_birth using `:comparison`

cpjmcquillan:
This replaces the custom validation with the default `validates` method using the `comparison` key.

Dependants and Applicants  with a date of birth in the future are invalid.

Tests have also been added.


jsugarman (approved):
🎖️ Looks good

PR #1153, opened 2022-10-25
EL-400: Allow 'unspecified_source_income'/'quarterly' irregular payment type

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/EL-400)

We want to capture income that is from an unspecified source, expressed by the user as a single total figure received over the past 3 months. By 'unspecified source' we mean income that doesn't fall under any of the named types of regular or irregular payments that CFE accepts, with no requirement on the user to describe the source in any way.

So this PR adds 'unspecified_source_income' as a valid irregular type, and 'quarterly' as a valid income frequency, to the irregular payments model/endpoint.

Also, because prior to this there was only one irregular payment type - student loans - there was a bit of logic that assumed all irregular payments were student loans, an assumption that started to break with the new type. So I unpicked that logic.

## Checklist

### DIFF
--- a/app/services/collators/gross_income_collator.rb
+++ b/app/services/collators/gross_income_collator.rb
@@ -41,11 +41,11 @@ module Collators
       # setup initial values here, populate_income_attrs above adds the other income(s)
       # to the default start values here
       {
-        total_gross_income: monthly_student_loan + monthly_unspecified_source_income,
+        total_gross_income: monthly_student_loan + monthly_unspecified_source,
         monthly_student_loan:,
-        monthly_unspecified_source_income:,
+        monthly_unspecified_source:,
         student_loan: categorised_income[:student_loan],
-        unspecified_source_income: categorised_income[:unspecified_source_income],
+        unspecified_source: categorised_income[:unspecified_source],
         monthly_other_income: categorised_income[:total],
         monthly_state_benefits:,
       }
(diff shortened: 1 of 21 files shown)


willc-work (approved):
LGTM - tested against an EFE instance

app/services/collators/gross_income_collator.rb
```diff
@@ -60,17 +62,46 @@ def monthly_student_loan
...
+    def irregular_payment_divisor(irregular_income_payment)
+      case irregular_income_payment.frequency
+      when CFEConstants::ANNUAL_FREQUENCY
+        12
+      when CFEConstants::QUARTERLY_FREQUENCY
+        3
+      else
+        raise "Unprocessable irregular income payment frequency: #{irregular_income_payment.frequency}"
+      end
+    end
```
cpjmcquillan, line 103: If we validate the payment frequency at the model level and we specify valid values in the schema, do we need to worry about raising specific errors for invalid frequencies?

I assume our dear friend `SonarCloud` complains about case statements without a default, but maybe we could fetch values from a hash instead?

```ruby
IRREGULAR_PAYMENT_PERIODS = {
  CFEConstants::ANNUAL_FREQUENCY => 12,
  CFEConstants::QUARTERLY_FREQUENCY => 3,
}.freeze

def irregular_payment_divisor(payment)
  IRREGULAR_PAYMENT_PERIODS.fetch(payment)
end
```

I wonder whether it might be helpful to define a method on the irregular payment model, or even break the normalisation calculation into its own class?

cpjmcquillan (approved):
I added some non-blocking comments and suggestions, but otherwise the change looks good to me.

One thing I wondered was whether the naming of the irregular payment types should be more consistent now there is more than one type?

Could we omit the `_income` from `unspecified_source_income`, or is it conveying some useful meaning?

Equally, would it be clearer if we added `_income` to `student_loan`?

patrick-laa:
@cpjmcquillan I love all of these suggestions, and will adopt. Let's go with `unspecified_source` for the new field rather than changing the student loan field. Thank you.

app/services/collators/gross_income_collator.rb
patrick-laa: I've added a `#monthly_equivalent_amount` method to the IrregularIncomePayment class, on the grounds that "monthly equivalent" seems like a fairly commonly used term and it's such a simple bit of logic that it feels like a reasonably thing for the model to be able to know and communicate about itself.

patrick-laa:
@cpjmcquillan I've made the changes you recommended; however I also appear to have committed the unforgivable offence of pushing twice in quick succession, and now the `deploy_uat` CircleCI step keeps failing because it wrongly thinks "another operation (install/upgrade/rollback) is in progress". I don't think I have the permissions to fix this by deleting the blocked release via helm CLI. Is there any chance you could do the honours?

cpjmcquillan:
@patrick-laa looks like you got it resolved!

Let me know if you still need me to do anything 👍

PR #1156, opened 2022-10-26
EL-412: Adopt TAP cucumber integration

patrick-laa:
[Link to story](https://dsdmoj.atlassian.net/browse/EL-412).

TAP recommended switching the "non-developer-friendly" integration tests from a Google Sheets/RSpec combo to Cucumber, and wrote a PoC to demonstrate that. EFE has decided to start documenting functionality we add to CFE through the new sort of integration tests rather than the old.

Therefore this PR brings the TAP PoC into the main codebase, with minimal updates to comply with the coding style guide and clarify the documentation. It also adds a cucumber test to document the new 'unspecified source' income functionality.


starswan (commented):
Looks fine to me

README.md
```diff
@@ -156,6 +156,18 @@ options:
 Each worksheet has an entry `Test Active` which can be either true or false.  If set to false, the worksheet will be skipped, unless it is
 the named worksheet using the `-w` command line switch.
 
+## Integration tests (cucumber)
+
+The `features` folder holds two sample cucumber tests that are a POC for a future project to replace the above spreadsheet-based ispec tests.
```
willc-work, line 161: should we remove reference to POC and state it is to run alongside the ispec tests for now, testing different features.

patrick-laa: Good point!

README.md
```diff
@@ -156,6 +156,51 @@ options:
...
+
+```
+bin/setup
+```
+
+
+In case the rails server does not start up, start it up to run the tests against them:
+
+```
+bin/rails server
```
jsugarman: this would start the development server. we should be running against the test database I think, no?!

```
 RAILS_ENV=test rails s
 ```

File: README.md [hunk 3]
```diff
@@ -156,6 +156,51 @@ options:
...
+```
+
+To run the tests
+
+```
+bundle exec cucumber
+```
+
+For convenience, we've also introduced git abstracted scripts for Windows based non-dev users. These can be found in the features folder. You will need to setup the `CUCUMBER_REPO_PATH` environment variable to get these scripts working. The name of the file describes the interaction it will perform. For example, save.bat is short for creating a branch, committing and pushing to Github repo.
```
jsugarman, line 169: 🤔 as discussed, I am not sure whether this is a step too far or not. Do we want non-devs pushing up branches?

features/support/env.rb
```diff
@@ -0,0 +1,77 @@
...
+ENV["RAILS_ENV"] ||= "test"
+ENV["APP_ENV"] ||= "test"
+
+# require 'cucumber/rails'
+require 'capybara/cucumber'
+require 'capybara/mechanize'
+
+# frozen_string_literal: true
+
+Capybara.app_host = "http://127.0.0.1:3000"
```
jsugarman: As discussed, this setup requires the rails server to be running and the test will use that rails server and whatever DB it is pointing at too. This is unusual in the  world of cucumber, unless I am missing something. 

This means it could be running against a local development database (or worse, if you run `bundle exec cucumber` on production 🤔 )

It is not truncating the test data either. I have tested this and it is creating 3 assessments for each run.

```
 # run test rails server instead
  RAILS_ENV=test rails s
 ```
 
We could, instead, be using hooks to run a test rails server on a specific port in a before_suite for example, and using that so that it is 
a. transparently using test database 
b. enable database cleaning to prevent leaking data between tests
c. not using the typical host and port that a dev server might

File: features/support/env.rb [hunk 5]
```diff
@@ -0,0 +1,77 @@
...
+Capybara.register_driver :mechanize do |app|
+  Capybara::Mechanize::Driver.new(proc {})
+end
+
+Capybara.default_driver = :mechanize
+
+# Remove/comment out the lines below if your app doesn't have a database.
+# For some databases (like MongoDB and CouchDB) you may need to use :truncation instead.
+# begin
+#   DatabaseCleaner.strategy = :transaction
```
jsugarman, line 53: once setup to use test databases only, this can be uncommented to prevent leakage between scenarios

jsugarman (commented):
@patrick-laa further to discussion I have left a few comments below. 

I would like to see feature tests that can be extended/added-to by BA's and testers but I am pretty sure that we need the setup to ensure it is using the test database only.

colinbruce:
Is the aim of the cucumber tests to act as a smoke test or simply run as features?
I've never seen cucumber require a pre-running server (although in fairness I have never configured it, it's always just been working!) and it seems like a non-standard setup will make this harder for a non-dev to add features

patrick-laa:
@colinbruce I agree, and having taken some soundings from other developers I think this is the consensus view. I will adapt the TAP spike to use a more standard cucumber approach!

jsugarman:
I was not keen on ispec originally but as a tool for BA's, who love spreadsheets 😄, and with a bit of a digging, I don't mind it. 

I think there is a place for more typical ruby/rails integration testing regardless. Whether we use cucumber or rspec feature tests is another matter, but cucumber at least opens the possibility for non devs to write features for devs to "wire-up"

Will pull changes and have a play before commenting further.

jsugarman:
> Will pull changes and have a play before commenting further.

changes look fine to me. be good to have other CFE devs review but 👍 from me

README.md
```diff
@@ -156,6 +156,16 @@ options:
 Each worksheet has an entry `Test Active` which can be either true or false.  If set to false, the worksheet will be skipped, unless it is
 the named worksheet using the `-w` command line switch.
 
+## Integration tests (cucumber)
+
+We are beginning the process of transitioning from "ispec" to cucumber. During the transition we have tests defined in both formats. Newer tests are written as cucumber features in the `features` folder.
```
jsugarman, line 161: we cannot get rid of ispec without approval from Dave's input would say as he has used them a lot to test expected values in the past

jsugarman (approved):
👍 In terms of adding feature tests. 

However this is baring comments regarding removal of ispec or cucumber usability by non-devs, over which I have raised questions.

README.md
patrick-laa: Fair point. I will reword the readme to remove the suggestion that getting rid of ispec is a definite.

naseberry:
The Eligibility team has forked the project: https://github.com/ministryofjustice/check-financial-eligibility-partner

They have since merged this into their fork, along with a bunch of cucumber tests covering our new functionality.

@patrick-laa has confirmed that this can be deleted.

PR #1162, opened 2022-10-27
Update applicant schema

cpjmcquillan:
Both `involvement_type` and `self_employed` fields were omitted from the
applicant schema.

The `involvement_type` field has been defined as a required enum field,
that can take a value of `"applicant"` or `null`.

The `self_employed` field has been defined as an optional boolean field.

The swagger documentation for the applicant endpoint has also been updated to
reflect the schema.

### DIFF
--- a/public/schemas/applicant_v5.json.erb
+++ b/public/schemas/applicant_v5.json.erb
@@ -10,6 +10,7 @@
       "required": [
         "date_of_birth",
         "has_partner_opponent",
+        "involvement_type",
         "receives_qualifying_benefit"
       ],
       "properties": {
--- a/public/schemas/applicant_v5.json.erb
+++ b/public/schemas/applicant_v5.json.erb
@@ -7,7 +7,11 @@
   "properties": {
     "applicant": {
       "type": "object",
-      "required": ["date_of_birth", "has_partner_opponent", "receives_qualifying_benefit"],
+      "required": [
+        "date_of_birth",
+        "has_partner_opponent",
+        "receives_qualifying_benefit"
+      ],
       "properties": {
         "date_of_birth": {
           "$ref": "<%= "file://#{@schema_dir}/common.json#date" %>"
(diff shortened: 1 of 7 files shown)


public/schemas/applicant_v5.json.erb
```diff
@@ -15,8 +15,15 @@
         "has_partner_opponent": {
           "type": "boolean"
         },
+        "involvement_type": {
+          "type": ["string", "null"],
+          "enum": ["applicant", "nil"]
```
cpjmcquillan, line 20: This feels weird to me.

I'd assume non-required fields would be allowed to be null.

Should we add optional fields to the schema at all?

public/schemas/applicant_v5.json.erb
```diff
@@ -15,8 +15,15 @@
         "has_partner_opponent": {
           "type": "boolean"
         },
+        "involvement_type": {
+          "type": ["string", "null"],
+          "enum": ["applicant", null]
+        },
```
cpjmcquillan, line 21: This feels weird to me.

I'd assume non-required fields would be allowed to be null.

Should we add optional fields to the schema at all?

colinbruce: another weird thing... IMO a schema defines what we will accept. How were we accepting submissions with extra fields?!

colinbruce (approved):

public/schemas/applicant_v5.json.erb
cpjmcquillan: That's what I would expect too.

It looks like the JSON validation isn't doing what we think it's doing (or at least what the code tells me it's doing).

When we [submit values that aren't defined in the schema](https://sentry.io/organizations/ministryofjustice/issues/3702754236/?project=5416052), it passes validation and attempts to create the records —ultimately [raising an unknown attribute error](https://sentry.io/organizations/ministryofjustice/issues/3702754301/?project=5404890).

I'll add a [task to the backlog to investigate](https://dsdmoj.atlassian.net/browse/AP-3583), because I think we should be making use of strict validation when it comes to the JSON schema.

cpjmcquillan: It gets even more confusing, because the swagger docs say `involvement_type` should be required 🤔

colinbruce:
Still looks good to me :+1: @jsugarman may spot something I missed though! 😄

jsugarman (approved):
Looks OK to me but have not tested against Apply.

Are we always sending involvement_type now?

cpjmcquillan:
Closing because I'm missing context now. If it's important I'm sure it will come up again

PR #1163, opened 2022-10-27
Employed applicants with dependants receive child care allowance

cpjmcquillan:
Before, employed applicants with child dependants were never able to receive the
child care cost allowance they were eligible for.

This was because the `#applicant_employed?` method in the childcare collator
always returned false.

Now that the method has been implemented, employed applicants that are eligible
will receive the correct child care cost allowance.

The applicant schema now defines an optional boolean employed attribute that
consumers can use to specify an applicant's employed status.


colinbruce (approved):
Codewise looks fine
Have you checked with Dave about the business logic? 
e.g. `becomes adult on` at 16 is that definitely when they stop receiving child benefits? I thought it was older if they were in full time education 🤷‍♂️

PR #1164, opened 2022-10-27
Define Faraday gem as an explicit dependency

cpjmcquillan:
Throughout the application, we are using the [Faraday gem] to make requests to the Legal Framework API.

Until now, we have been relying on Faraday being pulled in by another of our dependencies.

Ideally we would be using [Net::HTTP] since it comes as part of Ruby, but at the very least we should be explicit about the dependencies our application needs.

[Faraday gem]: https://github.com/lostisland/faraday
[Net::HTTP]: https://rubyapi.org/3.1/o/net/http

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,6 +20,8 @@ gem "puma", "~> 5.6"
 # Use ActiveStorage variant
 # gem 'mini_magick', '~> 4.8'
 
+gem "faraday", "~> 1.0"
+
 gem "sentry-rails", ">= 5.3.1"
 gem "sentry-ruby"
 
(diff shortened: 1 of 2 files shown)


colinbruce (approved):

naseberry:
Is there a reason for not choosing the latest version of the gem?

cpjmcquillan:
> Is there a reason for not choosing the latest version of the gem?

TLDR; Yes.

We're currently using Faraday in the application, despite not bundling the gem ourselves (i.e it's a transitive dependency).

It's pulled in by `googleauth`, which specifies `faraday (>= 0.17.3, < 2.0)`.

Since we have been relying on that version, I don't think it makes sense for us to move straight to a newer major version —even if we're now explicitly requiring the gem.

Ultimately we should be looking to remove the dependency completely, but I'm as yet unsure how much work that will take.

Does that make sense @naseberry?

PR #1165, opened 2022-10-27
EL-428: Acknowledge and account for SMOD assets

patrick-laa:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/EL-428)

'Subject matter of dispute' is an additional capital disregard that works somewhat similarly to the pensioner disregard. The requirement is for the disposable capital calculated result to disregard all assets that are marked as "subject matter of dispute", UP TO a threshold that is currently set at £100,000.00. Any disputed asset value above that threshold is included in the disposable capital calculation.

To that end, this PR adds the following:

- API additional optional fields to tell CFE when capital items, properties and vehicles are the subject matter of dispute
- Logic to disregard assets up to the SMOD disregard threshold from the disposable capital calculation
- SMOD disregard summary in result payload

### DIFF
--- a/app/services/calculators/subject_matter_of_dispute_disregard_calculator.rb
+++ b/app/services/calculators/subject_matter_of_dispute_disregard_calculator.rb
@@ -1,5 +1,7 @@
 module Calculators
   class SubjectMatterOfDisputeDisregardCalculator < BaseWorkflowService
+    delegate :disputed_capital_items, :disputed_vehicles, :disputed_properties, to: :capital_summary
+
     def value
       total_disputed_asset_value = disputed_capital_value +
         disputed_property_value +
@@ -19,15 +21,15 @@ module Calculators
     end
 
     def disputed_capital_value
-      capital_summary.capital_items.select(&:subject_matter_of_dispute).sum(&:value)
+      disputed_capital_items.sum(:value)
     end
 
     def disputed_property_value
-      capital_summary.properties.select(&:subject_matter_of_dispute).sum(&:assessed_equity)
+      disputed_properties.sum(:assessed_equity)
     end
 
     def disputed_vehicle_value
-      vehicles.select(&:subject_matter_of_dispute).sum(&:assessed_value)
+      disputed_vehicles.sum(:assessed_value)
     end
   end
 end
--- a/app/services/calculators/subject_matter_of_dispute_disregard_calculator.rb
+++ b/app/services/calculators/subject_matter_of_dispute_disregard_calculator.rb
@@ -1,0 +1,29 @@
+module Calculators
+  class SubjectMatterOfDisputeDisregardCalculator < BaseWorkflowService
+    def value
+      total_disputed_asset_value = disputed_capital_value +
+        disputed_property_value +
+        disputed_vehicle_value
+
+      [total_disputed_asset_value, threshold].min
+    end
+
+  private
+
+    def threshold
+      @threshold ||= Threshold.value_for(:subject_matter_of_dispute_disregard, at: submission_date) || 0
+    end
+
+    def disputed_capital_value
+      capital_summary.capital_items.select(&:subject_matter_of_dispute).sum(&:value)
+    end
+
+    def disputed_property_value
+      capital_summary.properties.select(&:subject_matter_of_dispute).sum(&:assessed_equity)
+    end
+
+    def disputed_vehicle_value
+      vehicles.select(&:subject_matter_of_dispute).sum(&:assessed_value)
+    end
+  end
+end
--- a/app/services/calculators/subject_matter_of_dispute_disregard_calculator.rb
+++ b/app/services/calculators/subject_matter_of_dispute_disregard_calculator.rb
@@ -5,13 +5,17 @@ module Calculators
         disputed_property_value +
         disputed_vehicle_value
 
-      [total_disputed_asset_value, threshold].min
+      if total_disputed_asset_value.positive? && threshold.nil?
+        raise "SMOD assets listed but no threshold data found for #{submission_date}"
+      end
+
+      [total_disputed_asset_value, threshold].compact.min
     end
 
   private
 
     def threshold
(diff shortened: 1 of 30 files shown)


db/migrate/20221027150350_add_smod_to_capital_items.rb
```diff
@@ -0,0 +1,8 @@
+class AddSmodToCapitalItems < ActiveRecord::Migration[7.0]
+  def change
+    add_column :capital_items, :subject_matter_of_dispute, :boolean, default: false
+    add_column :vehicles, :subject_matter_of_dispute, :boolean, default: false
+    add_column :properties, :subject_matter_of_dispute, :boolean, default: false
+    add_column :capital_summaries, :subject_matter_of_dispute_disregard, :decimal, default: "0.0", null: false
```
starswan, line 6: I think nullable: true is a good additional property to set here. Do we need default: false? (I can believe that the answer is yes, I'm just checking)

spec/factories/vehicle_factory.rb
```diff
@@ -5,5 +5,6 @@
     loan_amount_outstanding { Faker::Number.decimal(l_digits: 4, r_digits: 2) }
     date_of_purchase { Faker::Date.between(from: 6.years.ago, to: 2.months.ago) }
     in_regular_use { Faker::Boolean.boolean }
+    subject_matter_of_dispute { Faker::Boolean.boolean }
```
starswan, line 8: These feel too important to be set to a random value - should it not be false by default?

spec/integration/v5_full_assessment_spec.rb
```diff
@@ -244,11 +258,40 @@ def vehicle_params
           "loan_amount_outstanding" => "0.0",
           "date_of_purchase" => "2020-08-18",
           "in_regular_use" => true,
+          "subject_matter_of_dispute" => true,
```
starswan, line 261: I know it's probably outside of the scope of this change, but it worries me that this vehicle was purchased 2 months after the submission date in the test. Might it be sensible to do a separate 'test cleanup' PR to sort out this (and change the submission date to 2022)? WDYT?

spec/requests/swagger_docs/v5/capitals_spec.rb
```diff
@@ -32,7 +32,8 @@
                     bank_accounts: {
                       type: :array,
                       description: "One or more account details",
-                      example: [{ value: 1.01, description: "test name 1" }, { value: 100.01, description: "test name 2" }],
+                      example: [{ value: 1.01, description: "test name 1" },
```
starswan, line 35: should this first one be subject_matter_of_dispute: false?

app/services/calculators/subject_matter_of_dispute_disregard_calculator.rb
```diff
@@ -0,0 +1,29 @@
...
+        disputed_property_value +
+        disputed_vehicle_value
+
+      [total_disputed_asset_value, threshold].min
+    end
+
+  private
+
+    def threshold
+      @threshold ||= Threshold.value_for(:subject_matter_of_dispute_disregard, at: submission_date) || 0
```
starswan, line 14: IMHO we shouldn't default this to 0. If this ever happens, I would quite like a crash as the world has just ended (time has gone backwards)

patrick-laa: @starswan I don't entirely agree on this one. CFE has the functionality to retrospectively assess applications submitted in the past, which is why we have threshold data stretching back several years, and a client-defined submission date. My thought was that to be consistent (and avoid reconstructing what the SMOD upper limit was at various points in recent history), we should only enable SMOD considerations for the period for which we know what the threshold is. But a submission date reasonably far in the past is not a sign that the world has ended.

db/migrate/20221027150350_add_smod_to_capital_items.rb
patrick-laa: Fair point. The database is a mish-mash of user-provided data and calculated data. From the calculation perspective an asset is either treated as smod or it isn't, but a user can either tell us an asset is smod, or tell us it isn't, or not tell us at all. It's probably more valuable to store the user-provided information in the database and infer the calculation information from it.

spec/integration/v5_full_assessment_spec.rb
patrick-laa: Ah, but in this case the submission date in the test _is_ now 2022, so the date of purchase is in the past. But yes, I think a separate PR to tidy up the tests and deal with the age of the vehicle issue is probably a good thing.

patrick-laa:
@starswan two useful bits of information to throw into the mix here: Will C has confirmed that Apply can provide submission dates up to a year in the past, and Mike H has confirmed that the SMOD threshold hasn't changed in the past year. So for CFE to support SMOD, I think it's reasonable for us to add threshold data going a year back into the past and then raise an exception if a client requests to calculate SMOD-y eligibility for a date further back than that.

I also think that justifies updating the v5 integration test from a 2020 submission date to a 2022 submission date, because 2020 submission dates are no longer a realistic use case. (And I think there's a reasonable separate PR to improve how we do that test.

With that in mind, I think this is ready for re-review.

starswan (approved):

willc-work (approved):

colinbruce (approved):
Looks Good to me, one non-blocking observation :+1:

spec/services/creators/capitals_creator_spec.rb
```diff
@@ -12,6 +12,8 @@ module Creators
     let(:item1) { Faker::Appliance.equipment }
     let(:value1) { BigDecimal(Faker::Number.decimal(r_digits: 2), 2) }
     let(:value2) { BigDecimal(Faker::Number.decimal(r_digits: 2), 2) }
+    let(:smod1) { true }
```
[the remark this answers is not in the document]
patrick-laa: Good suggestion! I have added a new commit with the change. Thanks for the review - I think this is ready to be merged!

PR #1168, opened 2022-10-31
Fix employment without payments bug

cpjmcquillan:
An [existing bug resurfaced recently][sentry].

When an assessment has an employment but the employment has no employment payments, a `NoMethodError` is raised when calculating the range of monthly employment income payments in `Utilities::EmploymentIncomeVariationChecker`.

While this scenario is unlikely, it is possible —so we should account for it.

In the instance linked above, the applicant specified that they were not employed, but the HMRC response indicated that the applicant had an employment (albeit they had no employment payments).

This change ensures the most recent employment payment is only used for monthly employment values when there is at least one employment payment, and the range of employment income payments is below the threshold.

If there are no employment payments or the range exceeds the threshold, the blunt average of all employment payments is used. The blunt average will return `0` if there are no employment payments.

As part of this work, the `#calculate!` method has been simplified so employment records are only update once.

The associated specs have also been refactored, with the intention of testing the behaviour of the `#calculate!` method rather than focusing on implementation details.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3587)

[sentry]: https://sentry.io/organizations/ministryofjustice/issues/3345805354

### DIFF
--- a/app/models/employment.rb
+++ b/app/models/employment.rb
@@ -10,63 +10,50 @@ class Employment < ApplicationRecord
   validates :calculation_method, inclusion: { in: %w[blunt_average most_recent] }, allow_nil: true
 
   def calculate!
-    calculate_monthly_gross_income!
     Calculators::TaxNiRefundCalculator.call(self)
-    calculate_monthly_ni_tax!
-  end
 
-private
-
-  def calculate_monthly_gross_income!
-    if Utilities::EmploymentIncomeVariationChecker.new(self).below_threshold?
-      update!(
-        monthly_gross_income: employment_payments.order(:date).last.gross_income_monthly_equiv,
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
Code wise looks good to me and logic wise it appears to not be changing anything. 

I spent time checking the logic of having moved the `Calculators::TaxNiRefundCalculator.call(self)` to be the first call and it makes sense to me. Namely that would update positives to zero before further calculations are performed.

It looks like we have quite a few "integration tests" in the [employment spreadsheets here ](https://docs.google.com/spreadsheets/d/1fv0oNuQBIB_r2krsRCAM8lOVKdkaf8olbr6Wa-cVEE0/edit#gid=2058442362) that are all passing so 👍

PR #1169, opened 2022-10-31
Fix flaky integration spec

cpjmcquillan:
This fixes a [flaky spec].

The expectation asserted that the nested arrays in the response contained strings in a specific order, but this wasn't guaranteed to always be the case.

The order-agnostic `contain_exactly` matcher is preferred instead.

[flaky spec]: https://app.circleci.com/pipelines/github/ministryofjustice/check-financial-eligibility/6112/workflows/f7608ecd-e2ad-4232-8dc0-e350105b3aed/jobs/21133/tests


jsugarman (approved):
👍

PR #1190, opened 2022-11-16
Fix: Remove yarn.lock

colinbruce:
## What

This was re-introduced via a rebase in the removal of webpacker pull-request. It would be ignored but is superfluous so should be removed

## Checklist


naseberry (approved):

naseberry:
Not so superfluous as Github securities would continue to scan this file and therefore may create dependabot PR's for vulnerabilities

PR #1200, opened 2022-11-24
AP-3657: Update Section 8 proceeding type values

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3657)

Extend the accepted proceeding type ccms_code list in CFEConstants

Also removes the outdated ProceedingTypeThreshold class.  It was marked as removable once V4 had been removed.
All references to its valid_ccms_codes method have been replaced with calls direct to the CFEConstants version and a single test was calling the obsolete class to set a value of 99_999_999 which was already being set by the factory.  The call was superfluous so I have removed it

## Checklist


kmahern (approved):
lgtm

PR #1234, opened 2022-12-19
Only eager load Ruby code when running tests in CI

cpjmcquillan:
Before, the code was eager loaded every time a test was run.

When you're running a single test, this doesn't make much sense.

This updates the test configuration to eager load only when the `CI` environment
variable is set (i.e default Rails behaviour).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,8 +16,8 @@ executors:
     docker:
       - image: cimg/ruby:3.1.3-node
         environment:
-          - RAILS_ENV=test
-          - TZ: "Europe/London"
+          RAILS_ENV: test
+          TZ: "Europe/London"
   notification-executor:
     docker:
       - image: 'cibuilds/base:latest'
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -28,11 +29,12 @@ executors:
...
-          - PGHOST=localhost
-          - PGUSER=user
-          - TZ: "Europe/London"
-          - ALLOW_FUTURE_SUBMISSION_DATE: true
+          RAILS_ENV: test
+          PGHOST: localhost
+          PGUSER: user
+          TZ: "Europe/London"
+          ALLOW_FUTURE_SUBMISSION_DATE: true
+          CI: true
```
jsugarman, line 37: As per LFA
> is this needed? most (circlci, gitlab, travis from a quick google) CI/CD platforms set this anyway.

jsugarman (approved):
LGTM 👍

just a point about whether we need/should-be supplying the CI var explicitly

colinbruce:
Does this [Jira ticket](https://dsdmoj.atlassian.net/browse/AP-3107) match your expectations and is it something we should be concerned about?

cpjmcquillan:
> Does this [Jira ticket](https://dsdmoj.atlassian.net/browse/AP-3107) match your expectations and is it something we should be concerned about?

I opened these PRs off the back of trying to close off those tickets in JIRA.

What Stephen wrote in the description doesn't quite make sense. I think he meant to say that if `config.eager_load = false` then files that aren't loaded by any tests aren't included in coverage statistics.

PR #606, opened 2021-09-22
AP-2179 Get details of which thresholds are waived from LFA

stephenrichards:
## Get details of which thresholds are waived from LFA

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2179)

- Implemented `LegalFrameworkAPI::QueryService` to query LFA to get details of what thresholds are waived for specified proceeding types
- Amended `ProceedingTypeThreshold` model to query this service rather than use hard coded values

## Checklist

Before you ask people to review this PR:

- Tests and rubocop should be passing: `bundle exec rake`
- Github should not be reporting conflicts; you should have recently run `git rebase master`.
- There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- The PR description should say what you changed and why, with a link to the JIRA story.
- You should have looked at the diff against master and ensured that nothing unexpected is included in your changes.
- You should have checked that the commit messages say why the change was made.

PR #1253, opened 2023-01-04
Remove LFA query service

cpjmcquillan:
A TODO comment was added in #1005 indicating that this service object should be removed once v4 of the API was deprecated.

In #1047 all code for v3 and v4 was removed, but this was left behind.

This removes the redundant class and specs.


jsugarman (commented):
🧹 It certainly looks like it is not used anymore.

I do wonder whether the specs that were using the `mock_lfa_response` are excercising functionaliltiy that is now simply not requiring that data or whether there are redundant tests in there? Have you dug into that?

cpjmcquillan:
> I do wonder whether the specs that were using the mock_lfa_response are excercising functionaliltiy that is now simply not requiring that data or whether there are redundant tests in there? Have you dug into that?

The original intention of `mock_lfa_responses` was to ensure tests passed without needing to interact with LFA. 

Before #606, the `waived_threshold?` and `matter_type` information was defined in constants.

None of the tests are expecting the mocked data anymore, the code just hadn't been removed.

jsugarman (approved):

PR #1306, opened 2023-02-02
Revert "AP-3786 add nullable currency field type"

RoseSAK:
Reverts ministryofjustice/check-financial-eligibility#1304


jsugarman (approved):
What was the problem with the original
 change?

PR #1311, opened 2023-02-06
Fix: Update circle build-and-push job

colinbruce:
## What

Use the cloud platform executor and tidy the job
e.g. packages are not required

## Checklist


jsugarman (approved):
👍  Looks good

Shall we get rid of the basic-executor definition (line 7) while we are at it. And any other now redundant/unused bits like `setup_aws_login`? but either way

PR #1319, opened 2023-02-10
Fix: MonthlyEquivalentCalculator bug

colinbruce:
## What

The EFE team have identified an edge case where the frequency changing causes the calculator to return
incorrect results

## Checklist

### DIFF
--- a/app/models/concerns/monthly_equivalent_calculator.rb
+++ b/app/models/concerns/monthly_equivalent_calculator.rb
@@ -17,6 +17,7 @@ module MonthlyEquivalentCalculator
 
   def calculate_monthly_equivalent(collection:, date_method: :payment_date, amount_method: :amount)
     @converter = nil # reset the converter each time it's used
+    @frequency = nil
     return 0.0 if collection.empty?
 
     @monthly_equivalent_calculator_collection = collection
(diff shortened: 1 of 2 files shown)


app/models/concerns/monthly_equivalent_calculator.rb
```diff
@@ -17,6 +17,7 @@ def calculate_monthly_equivalent!(target_field:, collection:, date_method: :paym
 
   def calculate_monthly_equivalent(collection:, date_method: :payment_date, amount_method: :amount)
     @converter = nil # reset the converter each time it's used
+    @frequency = nil
```
cpjmcquillan, line 20: Should we be memoizing `converter` and `frequency` at all?

cpjmcquillan (approved):
I added a comment for us to think about.

Caching is hard, and a frequent source of bugs generally. If we weren't caching the values, would this bug be possible?

app/models/concerns/monthly_equivalent_calculator.rb
colinbruce: That can be one for the Eligibility Platform team to consider! 😄

starswan: Converter has to be memoized due to the way the implementation is structured (the memoization is called once to do it, and again for the answer). The structure of this code in CFE-Civil has changed so that local variables are used - which is how we found the bug. We had an new test that accidentally relied on the bug, which then broke when we accidentally fixed it

PR #1353, opened 2023-03-29
Update dependant allowances

cpjmcquillan:
From 10th April 2023, changes to means assessment fixed allowances for dependants will go live.

This updates the dependant allowances to the new amounts. All other thresholds have been left unchanged.

There is an allowance for dependant partners, but this hasn't been captured since we expect to integrate with the new `CFE-Civil` service before releasing partner means functionality in Apply.

These allowances should be applied for applications with submission dates on or after 2023-04-10.

Apply handles the logic surrounding application submission dates and delegated functions.

The Apply service passes the submission date when making requests to the assessment endpoint, so we do not need to worry about calculating it for an application.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3950)

### DIFF
--- a/config/thresholds/2023-04-10.yml
+++ b/config/thresholds/2023-04-10.yml
@@ -1,0 +1,68 @@
+---
+gross_income_upper: 2657
+infinite_gross_income_upper: 999_999_999_999
+dependant_step: 222
+dependant_increase_starts_after: 4
+disposable_income_lower: 315
+disposable_income_upper: 733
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
👍 can't speak to the accuracy of the other data in the thresholds but it makes sense and LGTM.

Have you discussed/considered adding ispec/spreadsheet-based tests for this?

